Issue #195, opened 2024-12-20
`TrajectoryFileCallback.after_transition` blows up if trajectory blows up before `before_transition`


jamesbraza:
As of `ldp==0.17.0`, if the trajectory blows up before `before_transition` gets called (e.g. during `agent.init_state`), then `after_transition` will blow up with an assertion error.

PR #196, opened 2024-12-20
Removed unnecessary `before_transition` from `TrajectoryFileCallback`

jamesbraza:
Closes https://github.com/Future-House/ldp/issues/195

### DIFF
--- a/ldp/alg/callbacks.py
+++ b/ldp/alg/callbacks.py
@@ -150,13 +150,8 @@ class TrajectoryFileCallback(Callback):
         """Create the filename for the output file."""
         return f"{traj_id}.jsonl"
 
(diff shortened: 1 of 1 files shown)


ldp/alg/callbacks.py
```diff
@@ -150,24 +150,15 @@ def _make_filename(self, traj_id: str, env: Environment) -> str:
         """Create the filename for the output file."""
         return f"{traj_id}.jsonl"
 
-    async def before_transition(
```
jamesbraza, line 153: We don't need to populate `self.out_files` up front, we can JIT populate within `after_transition`

sidnarayanan (approved):

PR #197, opened 2024-12-30
Suppress env frame errors; serialize logprob

sidnarayanan:
1. `TrajectoryFileCallback` changes assumed that all envs implemented `export_frame()`; this loosens that assumption
2. Serializing logprobs when dumping `OpResult`s

### DIFF
--- a/ldp/data_structures.py
+++ b/ldp/data_structures.py
@@ -123,7 +123,7 @@ class Trajectory(BaseModel):
             traj = cls(traj_id=json.loads(next(reader)))
             for json_line in reader:
                 data = json.loads(json_line)
-                # logprob may have been serialized, but cnanot be passed to
+                # logprob may have been serialized, but cannot be passed to
                 # OpResult, so remove it here.
                 with suppress(KeyError):
                     data["action"].pop("logprob")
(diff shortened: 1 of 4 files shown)


ldp/data_structures.py
```diff
@@ -121,7 +122,12 @@ def from_jsonl(cls, filename: str | os.PathLike) -> Self:
             reader = iter(f)
             traj = cls(traj_id=json.loads(next(reader)))
             for json_line in reader:
-                traj.steps.append(Transition(**json.loads(json_line)))
+                data = json.loads(json_line)
+                # logprob may have been serialized, but cnanot be passed to
+                # OpResult, so remove it here.
+                with suppress(KeyError):
+                    data["action"].pop("logprob")
```
jamesbraza, line 129: Why can `logprob` not be passed to `OpResult`'s action?

Also, typo in "cannot"

jamesbraza (approved):
Nice work, one less annoyance in the system (`export_frame` issues)

ldp/data_structures.py
sidnarayanan: `logprob` is a property that we extract from the OpCtx, not an attribute of the OpResult, so there's no way to set it on OpResult deserialization - the correct thing to do is also hydrate the OpCtx from the backend.

PR #198, opened 2024-12-30
`Agent.get_as` as an alternative to `get_asv`

jamesbraza:
I wanted to see what this `Agent.get_as` looks like since we've discussed it before. This PR's motivation is:
- Right now no agent in LDP estimate value
- Most unit tests (aka simple scripts) discard value

Pros:
- Simplicity/one less piece of information to comprehend for basic `Agent` implementations

Cons:
- Introduces room for ambiguity in runner code
- Introduces possible necessity for duplication across `get_as` and `get_asv`
- `get_as` is confusing because "as" is actually an English word, whereas "asv" is not
    - If you read as "get as" instead of "get action state" at first glance it's confusing

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ An agent uses tools in response to observations, which are just natural language
 ```py
 agent_state = await agent.init_state(tools=tools)
 new_action, new_agent_state, value = await agent.get_asv(agent_state, obs)
+# or if you don't need value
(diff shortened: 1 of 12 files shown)


jamesbraza:
Open to others opinions here, but reading through I feel the cons of `Agent.get_as` outweigh the pros of `Agent.get_asv`, so going to close this PR

PR #200, opened 2024-12-30
Got `SimpleAgent` example in the `README` to work

jamesbraza:


README.md
```diff
@@ -53,32 +53,38 @@ You can just emit actions directly if you want:
...
 
 class SimpleAgent(Agent):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.llm_call_op = LLMCallOp()
 
     async def init_state(self, tools):
         return AgentState([], tools)
 
-    @compute_graph()
```
jamesbraza, line 78: We don't need this for a minimal example

README.md
```diff
@@ -53,32 +53,38 @@ You can just emit actions directly if you want:
...
-    def __init__(messages, tools):
+    def __init__(self, messages, tools):
         self.messages = messages
         self.tools = tools
 
 
 class SimpleAgent(Agent):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.llm_call_op = LLMCallOp()
```
jamesbraza, line 78: We were missing this `llm_call_op` attribute

README.md
```diff
@@ -53,32 +53,38 @@ You can just emit actions directly if you want:
...
 
 ```py
-from ldp.graph import compute_graph
+from ldp.agent import Agent
+from ldp.graph import LLMCallOp
 
 
 class AgentState:
-    def __init__(messages, tools):
+    def __init__(self, messages, tools):
```
jamesbraza, line 70: We were missing `self`

README.md
```diff
@@ -53,32 +53,38 @@ You can just emit actions directly if you want:
 from aviary.core import ToolCall
 
 
-def get_asv(self, agent_state, obs):
+def get_asv(agent_state, obs):
```
jamesbraza, line 56: `self` is not needed for a free function `get_asv`

README.md
sidnarayanan: Can you change [this](https://github.com/Future-House/ldp/blob/main/ldp/alg/rollout.py#L303) to `OpResult.unwrap_value(action)` then? We are assuming the return of `get_asv` is `OpResult[ToolRequestMessage]`

sidnarayanan: Oops ignore me - forgot that an Op will make a compute graph if not already in one.

sidnarayanan (approved):

README.md
jamesbraza: > forgot that an Op will make a compute graph if not already in one.

Ah I see, and dang. In our other `Agent`s we avoid manipulating the graph's `LLMCallOp`'s prior inputs with this:

```python
next_state.messages = [*next_state.messages, result.value]
```

However, this `README`'s `SimpleAgent` doesn't have that, and I was trying to avoid that potential bug by removing the `compute_graph` part. But I guess it's unavoidable, so for now we can kick this can down the road

jamesbraza: If you have any ideas here, lmk

PR #204, opened 2025-01-13
Consolidating `StoreTrajectoriesCallback` with `TrajectoryFileCallback`, `async` file I/O, controlling env dump

jamesbraza:
This PR:
- Consolidates `TrajectoryFileCallback` into `StoreTrajectoriesCallback` with optional files
- Allows for control over `Environment.Frame` dump
- Dumps all files in `async` via `aiofiles`, resolving a TODO comment

### DIFF
--- a/ldp/alg/callbacks.py
+++ b/ldp/alg/callbacks.py
@@ -162,8 +162,7 @@ class TrajectoryFileCallback(Callback):
 
         traj = self.trajs[traj_id]
         traj.steps.append(transition)
-        # TODO: make this async?
-        traj.to_jsonl(self.out_files[traj_id])
+        await traj.to_jsonl(self.out_files[traj_id])
         if transition.done:
             with (
                 # Do not fail if the environment didn't implement export_frame().
--- a/ldp/alg/callbacks.py
+++ b/ldp/alg/callbacks.py
@@ -134,7 +134,7 @@ class StoreTrajectoriesCallback(Callback):
             frame = env.export_frame()
         except NotImplementedError:
             return None  # Allow for envs that didn't implement export_frame()
-        return frame.model_dump_json(exclude={"state"}, indent=2)
+        return frame.model_dump_json(indent=2)
 
     def __init__(
         self,
--- a/ldp/alg/callbacks.py
+++ b/ldp/alg/callbacks.py
@@ -172,9 +172,9 @@ class StoreTrajectoriesCallback(Callback):
         traj.steps.append(transition)
         if self.output_dir is not None:
             await traj.to_jsonl(self.traj_files[traj_id])
-            with self.env_files[traj_id].open("w") as f:
+            async with aiofiles.open(self.env_files[traj_id], "w") as f:
                 if to_dump := self.serialize_env(env, transition):
-                    f.write(to_dump)
+                    await f.write(to_dump)
 
     async def after_train_step(self, trajectories: Sequence[Trajectory]) -> None:
         self.train_traj_ids.extend([t.traj_id for t in trajectories])
--- a/ldp/alg/callbacks.py
+++ b/ldp/alg/callbacks.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 import logging
 import os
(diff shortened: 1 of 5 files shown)


ldp/alg/callbacks.py
```diff
@@ -123,54 +122,64 @@ async def after_update(self) -> None:
...
-        self.output_dir.mkdir(parents=True, exist_ok=True)
+    @staticmethod
+    def default_serialize_env(env: Environment, transition: Transition) -> str | None:
+        if not transition.done:
+            return None
+        try:
+            frame = env.export_frame()
+        except NotImplementedError:
+            return None  # Allow for envs that didn't implement export_frame()
+        return frame.model_dump_json(exclude={"state"}, indent=2)
```
sidnarayanan, line 135: Per in-person convo: must we drop state here?

ldp/alg/callbacks.py
```diff
@@ -123,54 +122,64 @@ async def after_update(self) -> None:
...
 
-        self.out_files: dict[str, Path] = {}
+    def __init__(
+        self,
+        output_dir: str | os.PathLike | None = None,
```
sidnarayanan, line 139: I think we can force the user to pass `output_dir`. It looks like we need it to be optional to support the `StoreTrajectoriesCallback` functionality, but I don't think anyone is actually using it (just grepped the codebase) - let's deprecate it and simplify this callback

ldp/alg/callbacks.py
```diff
@@ -123,54 +122,64 @@ async def after_update(self) -> None:
...
-                # Do not fail if the environment didn't implement export_frame().
-                suppress(NotImplementedError),
-                Path(self.env_files[traj_id]).open("w") as f,
-            ):
-                f.write(env.export_frame().model_dump_json(exclude={"state"}, indent=2))
+        if self.output_dir is not None:
+            await traj.to_jsonl(self.traj_files[traj_id])
+            async with aiofiles.open(self.env_files[traj_id], "w") as f:
+                if to_dump := self.serialize_env(env, transition):
+                    await f.write(to_dump)
```
sidnarayanan, line 176: Minor optimization, but perhaps could be `asyncio.gather`'d?

ldp/alg/callbacks.py
jamesbraza: Yeah I just kept this around to be backwards compatible, if one wants to serialize state, one can specify a serialization function.

Tho I do agree it's undesirable enough that we should not `exclude={"state"}` by default. @whitead what do you think of not `exclude`ing `state`?

ldp/alg/callbacks.py
jamesbraza: So we use `StoreTrajectoriesCallback` in unit tests to in-memory store trajectories. I don't want unit tests to have to use `tmpfile`s when really the tests just want to temporarily house `Trajectory` in-memory.

I would rather have it be optional, because people may not always have a filesystem

sidnarayanan: I see your point re: tests, but:
> I would rather have it be optional, because people may not always have a filesystem
I would contend there isn't a use-case for this callback if you don't have a filesystem - all it'd do is aggregate things into a list. You'd have to subclass and add some further post-processing for any real utility. That would be fairly simple to write from scratch and I'm not sure complicating this class is worth saving the user a little boilerplate.

ldp/alg/callbacks.py
jamesbraza: Sure nice comment, just implemented this

ldp/alg/callbacks.py
sidnarayanan: If you don't want to add that in this PR, I can open a follow-up with the simplification / modified tests. IMO it's worth keeping core parts of the code simple and easy-to-read.

jamesbraza: The use case with no filesystem is basically storing parts of the trajectory for post-rollout processing (e.g. evaluating consensus sampling still within the same Python execution).

Since the file I/O can become sizable (especially in longer trajectories), requiring it forces one to incur a perf hit even though one just needs an in-memory cache.

Lastly, I like being able to not worry about or requiring filesystem for DIY scripts and quick unit tests.

---

Let's just leave this debate for the future

ldp/alg/callbacks.py
jamesbraza: Alright, I just implemented this

ldp/alg/callbacks.py
sidnarayanan: Idk, it still feels like we're trying to put two (three?) unrelated things in one callback: aggregating trajectories in a list for another consumer, serializing trajectories to JSONLs, serializing final environment frames to JSONs. 

Happy to defer to another PR, but I do want to keep this part of the codebase simple and easy to read.

jamesbraza:
I extracted some of the improvements into https://github.com/Future-House/ldp/pull/205

ldp/alg/callbacks.py
jamesbraza: Sure no problem, we can just revisit this merge later

PR #205, opened 2025-01-14
`TrajectoryFileCallback` having `async` file I/O and controlling `Environment` dump

jamesbraza:
This PR:
- Dumps all files in `async` via `aiofiles`, resolving a TODO comment
- Allows for control over `Environment.Frame` dump
- Makes attribute names more intuitive

### DIFF
--- a/ldp/alg/callbacks.py
+++ b/ldp/alg/callbacks.py
@@ -162,8 +162,7 @@ class TrajectoryFileCallback(Callback):
 
         traj = self.trajs[traj_id]
         traj.steps.append(transition)
-        # TODO: make this async?
-        traj.to_jsonl(self.out_files[traj_id])
+        await traj.to_jsonl(self.out_files[traj_id])
         if transition.done:
             with (
                 # Do not fail if the environment didn't implement export_frame().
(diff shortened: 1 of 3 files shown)


ldp/alg/callbacks.py
```diff
@@ -139,38 +139,58 @@ async def after_eval_step(self, trajectories: Sequence[Trajectory]) -> None:
...
+    @staticmethod
+    def default_serialize_env(env: Environment, transition: Transition) -> str | None:
+        """Export a JSON-serialized Frame if the transition is done."""
+        if not transition.done:
+            return None
+        try:
+            frame = env.export_frame()
+        except NotImplementedError:
+            return None  # Allow for envs that didn't implement export_frame()
+        return frame.model_dump_json(exclude={"state"}, indent=2)
```
jamesbraza, line 151: I decided to keep state exclusion for now, because on a test-set run with PaperQA Local, `Frame` exports from just the `done`/`not failed` trajectories takes 0.5 GB. With 10 replicates, ROM usage starts to add up quickly

sidnarayanan: This actually explains why my TS RAM blows up so dramatically - must be something in state.

sidnarayanan (approved):

PR #208, opened 2025-01-15
`remove_nonterminal_branches`

sidnarayanan:
A method to create a tree with only terminal branches

### DIFF
--- a/ldp/data_structures.py
+++ b/ldp/data_structures.py
@@ -337,7 +337,7 @@ class TransitionTree:
             # step.metadata["advantage"] = step.value - state_values[parent_id]
 
     def remove_nonterminal_branches(self) -> TransitionTree:
-        """Creates a new tree with only branches that end in terminal states."""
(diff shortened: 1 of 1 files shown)


ldp/data_structures.py
```diff
@@ -336,6 +336,25 @@ def compute_advantages(self) -> None:
             # See docstring for explanation.
             # step.metadata["advantage"] = step.value - state_values[parent_id]
 
+    def remove_nonterminal_branches(self) -> TransitionTree:
+        """Creates a new tree with only branches that end in terminal states."""
```
jamesbraza, line 340: Can you add a comment stating failure is not considered here

sidnarayanan: Clarified `done=True`. Technically one could write a pathological environment in which both `step.done` and `step.failed` are true, and we'd include that branch in this method. So I don't want to claim we exclude failures.

jamesbraza (approved):

PR #209, opened 2025-01-15
Add num_epochs to OfflineTrainer

sidnarayanan:
See title

### DIFF
--- a/ldp/alg/runners.py
+++ b/ldp/alg/runners.py
@@ -14,7 +14,7 @@ from ldp.agent import Agent
 from ldp.alg.optimizer import Optimizer
 from ldp.data_structures import Trajectory
 from ldp.graph import OpResult, eval_mode, train_mode
-from ldp.shims import tqdm, trange
+from ldp.shims import tqdm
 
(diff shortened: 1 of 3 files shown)


ldp/alg/runners.py
```diff
@@ -298,7 +298,7 @@ class OfflineTrainerConfig(BaseModel):
         ge=1,
     )
     clear_ctx_at_each_iter: bool = False
-    # TODO: add some concept of eval loops
```
sidnarayanan, line 301: Removed this comment since, with `num_epochs>1`, one can insert a callback to do eval.

ldp/alg/runners.py
```diff
@@ -298,7 +298,7 @@ class OfflineTrainerConfig(BaseModel):
...
+    num_epochs: int = 1
```
jamesbraza, line 301:
```suggestion
    num_epochs: int = Field(default=1, ge=0)
```

Just adding a nice validator for negative values

jamesbraza (approved):
Nice work

PR #211, opened 2025-01-16
Consensus sampling decomposition and per-group standard error

jamesbraza:
- Decomposes `evaluate_consensus` to operate per-group a free function
- Adds support for binomial standard error on a per-group basis
- Removed `num_samples` bad default of `1`
- Added more docs/code comments in the docstring


ldp/alg/algorithms.py
```diff
@@ -132,67 +133,111 @@ def gvizify(x: Any) -> str:
...
+        extract_answer_fn = cast(Callable[[TData], TAnswer], extract_answer_fn)
+        answers = [extract_answer_fn(x) for x in sampled]
+
+    # Compute consensus: mode of the sampled answers
+    most_common = collections.Counter(answers).most_common()
+    # NOTE: If there are multiple modes, just use the first one
+    consensus_answer, consensus_count = most_common[0]
+    sample_acc_mean = consensus_count / num_samples
+    # Use binomial standard error since we're comparing ideal vs not ideal
+    sample_acc_ste = np.sqrt(sample_acc_mean * (1.0 - sample_acc_mean) / num_samples)
```
ludomitch, line 237: I've been using wilson confidence interval which tends to perform better in a variety of edge cases that we will encounter quite often eg low sample size:

https://arc.net/l/quote/vcnpsgab

ldp/alg/algorithms.py
```diff
@@ -132,67 +133,111 @@ def gvizify(x: Any) -> str:
...
 
+    ideal_count = sum(await asyncio.gather(*itertools.starmap(index, groups.items())))
     return grouped_consensus, ideal_count / len(groups) if groups else 0.0
 
 
+async def evaluate_consensus(
+    data: Sequence[TData],
+    extract_answer_fn: Callable[[TData], TAnswer | Awaitable[TAnswer]],
+    num_samples: int,
+    seed: random.Random | int | None = None,
```
ludomitch, line 193: Why do we need to support seed being instantiated as random.Random?

ldp/alg/algorithms.py
ludomitch:
```
import pandas as pd
from statsmodels.stats.proportion import proportion_confint
def get_wilson_ci(row, confidence_level=0.95):
    successes = int(row['success'].sum())
    n = len(row['success'])
    ci_lower, ci_upper = proportion_confint(
        count=successes, 
        nobs=n,
        method='wilson',  # Wilson score interval
        alpha=1-confidence_level
    )
    return pd.Series({
        'success_rate': successes/n,
        'ci_lower': ci_lower,
        'ci_upper': ci_upper,
        'n_trials': n,
        'n_successes': successes
    })
```

ldp/alg/algorithms.py
```diff
@@ -132,67 +133,111 @@ def gvizify(x: Any) -> str:
...
+    """
+    if isinstance(seed, int):
+        rand: ModuleType | random.Random = random.Random(seed)
+    elif isinstance(seed, random.Random):
+        rand = seed
+    else:
+        rand = random
+
+    if len(data) < num_samples:  # Too few items, sample with replacement
+        sampled = [rand.choice(data) for _ in range(num_samples)]
```
ludomitch, line 217: numpy has the replace functionality in-built

```
    sampled_votes = np.random.choice(
        predictions, size=min(k, len(predictions)), replace=False
    )
```

ldp/alg/algorithms.py
jamesbraza: It's because if you look, `bulk_evaluate_consensus` makes a `Random` there. Without this I would have had to redo the test stubs (that relied on a certain seeding).

In general with integration testing or concurrency, having one `Random` object across everything is desirable.

If you think we don't need this, then I can change the test stubs and just accomodate `int | None`

ldp/alg/algorithms.py
jamesbraza: Hm maybe we should allow folks to pass their own `Callable` closure for measuring uncertainty. The closure lets one specify confidence interval like you have, it's a good idea.

Also an aside is `ldp` doesn't depend on `pandas`, but we can make an extra for this I guess.

ldp/alg/algorithms.py
jamesbraza: Sure we can go to `numpy` for this stuff, prolly more performant too

ldp/alg/algorithms.py
```diff
@@ -132,67 +133,111 @@ def gvizify(x: Any) -> str:
...
+    else:  # Sample without replacement
+        sampled = rand.sample(data, num_samples)
+
+    # Get answers for the sampled data
+    if is_coroutine_callable(extract_answer_fn):
```
sidnarayanan, line 222: See note here: https://github.com/Future-House/ldp/pull/212/files#r1920598727

jamesbraza: Fixed this separately in https://github.com/Future-House/ldp/pull/213

ldp/alg/algorithms.py
jamesbraza: Moved to this, resolving

ldp/alg/algorithms.py
jamesbraza: Added support for `numpy.random.Generator` too. Resolving this one as we also discussed in-person

ldp/alg/algorithms.py
jamesbraza: Alright, this callback has just been created 👌 resolving this one

ludomitch (approved):

sidnarayanan (commented):
Not approving just yet - I have a concern about the sampling-with-replacement part of the code.

ldp/alg/algorithms.py
```diff
@@ -126,73 +126,128 @@ def gvizify(x: Any) -> str:
...
+    rand = (
+        seed
+        if isinstance(seed, np.random.Generator | random.Random)
+        else np.random.default_rng(seed)
+    )
+
+    if len(data) < num_samples:  # Too few items, sample with replacement
+        sampled: Iterable[TData] = [rand.choice(data) for _ in range(num_samples)]  # type: ignore[arg-type]
+    elif isinstance(rand, random.Random):  # Built-in random sample with replacement
+        sampled = rand.sample(data, num_samples)
```
sidnarayanan, line 228: I don't think we should be supporting sampling with replacement at all. For one, the standard error you have above assumes IID trials. And this will increase the variance of the consensus answer.

What motivated supporting it even if `len(data) >= num_samples`?

ldp/alg/algorithms.py
```diff
@@ -126,73 +126,128 @@ def gvizify(x: Any) -> str:
...
+            consensus answer.
+    """
+    rand = (
+        seed
+        if isinstance(seed, np.random.Generator | random.Random)
+        else np.random.default_rng(seed)
+    )
+
+    if len(data) < num_samples:  # Too few items, sample with replacement
+        sampled: Iterable[TData] = [rand.choice(data) for _ in range(num_samples)]  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
ludomitch: FWIW I agree. I think I commented somewhere that I don't see any use case for sampling with replacement to be a valid strategy for us.

ldp/alg/algorithms.py
```diff
@@ -126,73 +126,128 @@ def gvizify(x: Any) -> str:
...
+    elif isinstance(rand, random.Random):  # Built-in random sample with replacement
+        sampled = rand.sample(data, num_samples)
+    else:  # NumPy random sample with replacement
```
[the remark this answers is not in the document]
jamesbraza: Oh nice catch, when adding the NumPy RNG last night I had a copy/paste error here. Thanks for this

ldp/alg/algorithms.py
```diff
@@ -126,73 +126,128 @@ def gvizify(x: Any) -> str:
...
             compute accuracy with, or pass NO_IDEAL_ANSWER to skip this calculation.
-        num_samples: Number of samples to choose from the N total.
-        seed: Optional seed for sampling.
+        consensus_callback: SEE below evaluate_consensus.__doc__.
 
     Returns:
-        Two-tuple of consensus list generated by collections.Counter.most_common and
-            the proportion of groups for which the consensus matches the ideal.
+        Two-tuple of the consensus list generated by collections.Counter.most_common
+            and the proportion of groups for which the consensus matches the ideal.
```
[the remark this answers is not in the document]
jamesbraza: It was halfway right, but it failed to be explicit that "the consensus list" is really a mapping of group keys to consensus lists. I updated the docstring to make this clear

ldp/alg/algorithms.py
```diff
@@ -126,73 +126,128 @@ def gvizify(x: Any) -> str:
...
+            be async so this can be done using a LLM call.
+        num_samples: Number of samples to choose from the N total.
+        seed: Optional seed for sampling.
+        consensus_callback: Optional callback function called just after computing
+            consensus, it's passed the consensus answer, consensus size, and
+            sample size. This is useful for stuff like summary statistics.
+
+    Returns:
+        Two-tuple of consensus bins generated by collections.Counter.most_common and the
+            consensus answer.
```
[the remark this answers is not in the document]
jamesbraza: Great suggestion, implemented. Since I was the dev, I didn't realize this was a bit of an opaque description. Thanks for pointing out!

ldp/alg/algorithms.py
```diff
@@ -126,73 +126,128 @@ def gvizify(x: Any) -> str:
...
 NO_IDEAL_ANSWER_FN: Literal["NO_IDEAL_ANSWER_FN"] = "NO_IDEAL_ANSWER_FN"  # Sentinel
 
 
-async def evaluate_consensus(
+async def bulk_evaluate_consensus(
     data: Iterable[TData],
     grouping_fn: Callable[[TData], TGroupKey],
     extract_answer_fn: Callable[[TData], TAnswer | Awaitable[TAnswer]],
+    num_samples: int,
+    seed: int | None = None,
```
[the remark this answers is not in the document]
jamesbraza: Yeah definitely, just allowed it. Letting `seed` be a pass through actually simplifies the code a bit too 👌

ldp/alg/algorithms.py
```diff
@@ -126,73 +126,128 @@ def gvizify(x: Any) -> str:
...
-            extract_answer_fn = cast(Callable[[TData], TAnswer], extract_answer_fn)
-            answers = [extract_answer_fn(x) for x in sampled]
-
-        # Compute consensus: mode of the sampled answers
-        grouped_consensus[group_key] = collections.Counter(answers).most_common()
-        # NOTE: If there are multiple modes, just use the first one
-        consensus: TAnswer = grouped_consensus[group_key][0][0]
-        if ideal_answer_fn != NO_IDEAL_ANSWER_FN:
+
+    async def index(group_key: TGroupKey, group: list[TData]) -> int:
```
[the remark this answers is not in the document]
jamesbraza: Yep a valid point and agreed.

I had named it `index` basically because it's adding new entries into a `dict`, but you're right that:
- This name misaligns with its return type of matching the ideal answer
- `index` name is somewhat vague, regardless

I updated it accordingly, thanks

ldp/alg/algorithms.py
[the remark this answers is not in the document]
jamesbraza: Great catch, implemented, and included this rationale in the `ValueError`. Thanks

ldp/alg/algorithms.py
jamesbraza: Yeah this comment should have said `Built-in random sample without replacement` as `random.Random.sample ` is without replacement. But due to a copy-paste error I introduced during this PR it said `with replacement`, which I have now fixed. Sorry for confusion here

ldp/alg/algorithms.py
```diff
@@ -126,73 +126,128 @@ def gvizify(x: Any) -> str:
...
+        return 0
 
+    ideal_count = sum(await asyncio.gather(*itertools.starmap(index, groups.items())))
     return grouped_consensus, ideal_count / len(groups) if groups else 0.0
 
 
+async def evaluate_consensus(
+    data: Sequence[TData],
+    extract_answer_fn: Callable[[TData], TAnswer | Awaitable[TAnswer]],
+    num_samples: int,
```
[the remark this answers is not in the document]
jamesbraza: This is implemented now 👌

ldp/alg/algorithms.py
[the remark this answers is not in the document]
jamesbraza: Just added a test of this too

jamesbraza:
Gonna merge, @sidnarayanan feel free to make other comments and I will get back to them

PR #212, opened 2025-01-16
LLM response validation

sidnarayanan:
This PR implements user-specified response validation for `LLMCallOp` with retries. I've implemented a validator for mangled tool calls from Anthropic, but one could also use this for rejection sampling, etc. 

I toyed around with a couple other places to put this (`llmclient.LiteLLMModel`, `SimpleAgent`), but found this to be most natural and easiest to configure. Ideally I'd have liked to implement this as a `litellm` callback, but I don't think their retry mechanism is triggered by callback exceptions.

### DIFF
--- a/ldp/graph/common_ops.py
+++ b/ldp/graph/common_ops.py
@@ -307,7 +307,8 @@ class LLMCallOp(Op[Message]):
 
         @tenacity.retry(
             retry=tenacity.retry_if_exception_type(ResponseValidationError),
-            stop=tenacity.stop_after_attempt(num_retries),
+            # num_retries+1 because the first call is not a retry
+            stop=tenacity.stop_after_attempt(num_retries + 1),
             wait=tenacity.wait_fixed(1),
         )
         async def call_and_validate() -> LLMResult:
--- a/ldp/graph/common_ops.py
+++ b/ldp/graph/common_ops.py
@@ -214,6 +214,15 @@ class LLMCallOp(Op[Message]):
         num_samples_logprob_estimate: int = 0,
         response_validator: Callable[[LLMResult], Awaitable[None] | None] | None = None,
     ) -> None:
+        """Initializes the LLMCallOp.
+
+        Args:
+            num_samples_logprob_estimate: The number of samples used to estimate the partition
+                function at T!=1. Defaults to 0 (calculation is skipped).
+            response_validator: An optional callable (can be async) that validates the response.
+                It should raise an exception if the response is invalid. The Op will retry up
+                to `config.get('num_retries', 0)` times if validation fails.
+        """
         super().__init__()
         self.num_samples_partition_estimate = num_samples_logprob_estimate
         self.response_validator = response_validator
--- a/ldp/graph/common_ops.py
+++ b/ldp/graph/common_ops.py
@@ -11,6 +11,7 @@ from functools import lru_cache
 from typing import Generic, TypeVar, cast, overload
 
 import numpy as np
+import tenacity
 import tree
 from aviary.core import Message, Tool, ToolRequestMessage, is_coroutine_callable
 from llmclient import (
@@ -201,12 +202,21 @@ class PromptOp(FxnOp[str]):
         return super(FxnOp, self).__repr__()
 
 
+class ResponseValidationError(Exception):
+    """Raised when a response from the LLM does not pass user-specified validator."""
+
+
 class LLMCallOp(Op[Message]):
     """An operation for LLM calls interaction."""
 
-    def __init__(self, num_samples_logprob_estimate: int = 0) -> None:
+    def __init__(
+        self,
+        num_samples_logprob_estimate: int = 0,
+        response_validator: Callable[[LLMResult], bool | Awaitable[bool]] | None = None,
+    ) -> None:
         super().__init__()
         self.num_samples_partition_estimate = num_samples_logprob_estimate
+        self.response_validator = response_validator
 
     @overload
     async def forward(
@@ -253,8 +263,12 @@ class LLMCallOp(Op[Message]):
             # if no tools are provided, tool_choice must be 'none'
             tool_choice = "none"
 
-        result = await model.call_single(
-            messages=msgs, tools=tools, tool_choice=tool_choice
+        result = await self._call_single_and_maybe_validate(
+            model=model,
+            num_retries=config.get("num_retries", 0),
+            messages=msgs,
+            tools=tools,
+            tool_choice=tool_choice,
         )
         if result.messages is None:
             raise ValueError("No messages returned")
@@ -282,6 +296,34 @@ class LLMCallOp(Op[Message]):
         self.ctx.update(call_id, "logprob", logprob)
 
         return result.messages[0]
+
+    async def _call_single_and_maybe_validate(
+        self, model: LLMModel, num_retries: int, **kwargs
+    ) -> LLMResult:
+        if not self.response_validator:
+            # If a response validator is not supplied, then we should not do any retries here - leave
+            # that for LiteLLM to handle.
+            return await model.call_single(**kwargs)
+
+        @tenacity.retry(
+            retry=tenacity.retry_if_exception_type(ResponseValidationError),
+            stop=tenacity.stop_after_attempt(num_retries),
+            wait=tenacity.wait_fixed(1),
+        )
+        async def call_and_validate() -> LLMResult:
+            result = await model.call_single(**kwargs)
+
+            validated = cast(Callable, self.response_validator)(result)
+            if is_coroutine_callable(self.response_validator):
+                validated = await validated
+            if not validated:
+                # Don't provide more info here - user can control error logging
+                # in the validator if desired.
+                raise ResponseValidationError
+
+            return result
+
+        return await call_and_validate()
 
     async def compute_logprob(
         self,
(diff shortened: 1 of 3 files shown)


tests/test_ops.py
```diff
@@ -163,7 +164,7 @@ async def generate_story() -> str:
     async def test_empty_tools(self) -> None:
         llm_call_op = LLMCallOp()
         message_result = await llm_call_op(
-            LLMModel.model_fields["config"].default,
+            LLMModel.model_fields["config"].default_factory(),  # type: ignore[call-arg, misc]
```
sidnarayanan, line 167: I have no idea how this used to work. `model_fields["config"].default` is `PydanticUndefinedObject` , not a dictionary. 

That obviously breaks when this PR does `config.get("num_retries", 0)`, but I don't understand why `llmclient` didn't break on this before - it's expecting a dictionary.

ldp/graph/common_ops.py
```diff
@@ -201,12 +202,21 @@ def __repr__(self) -> str:
...
+
+
 class LLMCallOp(Op[Message]):
     """An operation for LLM calls interaction."""
 
-    def __init__(self, num_samples_logprob_estimate: int = 0) -> None:
+    def __init__(
+        self,
+        num_samples_logprob_estimate: int = 0,
+        response_validator: Callable[[LLMResult], bool | Awaitable[bool]] | None = None,
```
jamesbraza, line 215: If you want to do this, I think it would be `Callable[[LLMResult], bool] | Callable[[LLMResult], Awaitable[bool]] | None = None`

Unless I am wrong, I don't think `bool | Awaitable[bool]` is the same thing

ldp/graph/common_ops.py
```diff
@@ -283,6 +297,35 @@ async def forward(
...
+            # num_retries+1 because the first call is not a retry
+            stop=tenacity.stop_after_attempt(num_retries + 1),
+            wait=tenacity.wait_fixed(1),
+        )
+        async def call_and_validate() -> LLMResult:
+            result = await model.call_single(**kwargs)
+
+            validated = cast(Callable, self.response_validator)(result)
+            if is_coroutine_callable(self.response_validator):
+                validated = await validated
```
jamesbraza, line 319: So the design pattern for validators is:
- Return `None` if the validator passes
- Blows up if the validator fails

Think about how Pydantic `BaseModel` validations work. If they pass, the code silently proceeds with instantiation, otherwise a `ValidationError` is thrown

Can we somehow modify this code to be:

```python
try:
    self.response_validator(result)
except Exception as exc:
    raise ResponseValidationError() from exc
```

ldp/graph/common_ops.py
sidnarayanan: lol, I copied this from you: https://github.com/Future-House/ldp/blob/1ad551a429e295d0c80bce776eeca917442837b6/ldp/alg/algorithms.py#L138 . 

And actually, I think this syntax is correct, see here: https://peps.python.org/pep-0483/#covariance-and-contravariance. Specifically:
> [Callable] is covariant in the return type, but contravariant in the arguments.

tests/test_ops.py
jamesbraza: It's a regression from https://github.com/Future-House/ldp/pull/175 (cc @maykcaldas)
- Before: https://github.com/Future-House/ldp/blob/v0.15.0/ldp/llms/chat.py#L132-L137
- After: https://github.com/Future-House/llm-client/blob/v0.0.7/llmclient/llms.py#L658-L669

It went from `default` to `default_factory`

Your fix here is appropriate, it's too bad Pydantic validations didn't catch this earlier

ldp/graph/common_ops.py
jamesbraza: Okay gotchu and we discussed in person, resolving this one

tests/test_ops.py
sidnarayanan: Yeah that I understand, but I didn't understand how `LLMModel(config=PydanticUndefined)` worked - it should've crashed since we treat `config` as a dict.

And here is the answer:
image

jamesbraza: Made this: https://github.com/pydantic/pydantic/issues/11288

ldp/graph/common_ops.py
```diff
@@ -283,6 +297,35 @@ async def forward(
...
+            if not validated:
+                # Don't provide more info here - user can control error logging
+                # in the validator if desired.
+                raise ResponseValidationError
```
jamesbraza, line 323:
```suggestion
                raise ResponseValidationError(f"Validator {self.response_validator} failed.)
```

Let's put a message here because in log files what happens is this shows up as blank line, unless your log formatter also shows the exception type

ldp/graph/common_ops.py
jamesbraza: Okay we can just rethink this later, it's fine with me for now

ldp/graph/common_ops.py
```diff
@@ -283,6 +297,35 @@ async def forward(
...
+            return await model.call_single(**kwargs)
+
+        @tenacity.retry(
+            retry=tenacity.retry_if_exception_type(ResponseValidationError),
+            # num_retries+1 because the first call is not a retry
+            stop=tenacity.stop_after_attempt(num_retries + 1),
+            wait=tenacity.wait_fixed(1),
+        )
+        async def call_and_validate() -> LLMResult:
+            result = await model.call_single(**kwargs)
```
jamesbraza, line 315: Can you add a `# NOTE:` or something somewhere talking about how we can get many retries here:
- LiteLLM retrying some things
- `response_validator` leading to retrying other things

The max amount of retries can now be `num_retries ** 2`, right?

ldp/graph/common_ops.py
sidnarayanan: Alright, when doing this I saw the light - it's nice to be able to chain exceptions and see the full traceback (RetryError -> ResponseValidationError -> original exception). I've switched to the pattern you suggested here: https://github.com/Future-House/ldp/pull/212#discussion_r1917863172

ldp/graph/common_ops.py
```diff
@@ -283,6 +306,39 @@ async def forward(
...
+
+            try:
+                validated = cast(Callable, self.response_validator)(result)
+                if inspect.isawaitable(validated):
```
sidnarayanan, line 331: Note we need to use `isawaitable` not `is_coroutine_callable` to properly support `Callable[[...], X | Awaitable[X]]` here (specifically the intersection in the Venn diagram would've been missed by the latter). cc @jamesbraza

jamesbraza: Should we modify `is_coroutine_callable`? Also, do you have test coverage for what led you to figure out this issue?

sidnarayanan: No test - just noticed the logic: `Callable[..., X | Awaitable[X]]` captures `def f(...) -> X`, `async def f(...) -> X` and `def f(...) -> X | Awaitable[X]`. `is_coroutine_callable` wouldn't catch the last. And I don't think that's the purpose of that function, so need to modify.

ldp/graph/common_ops.py
```diff
@@ -355,6 +411,27 @@ def get_examples(self) -> list[tuple[LLMResult, float]]:
...
+
+        for msg in response.messages:
+            if not isinstance(msg, ToolRequestMessage):
+                continue
+
+            for tc in msg.tool_calls:
+                if any(x in tc.function.name for x in "{}<>()"):
+                    err_msg = f"Found bad tool call: {tc}"
+                    logger.error(err_msg)
+                    raise ValueError(err_msg)
```
jamesbraza, line 433: If it's a non-critical error, it should be logged with `warning` level, otherwise clients have to make custom filters to get clean logs in the face of flaky LLMs.

Secondly, if to get logging across `tenacity` retries, the route to go is with tenacity's `before_sleep_log(logger, logging.WARNING)`

tests/test_ops.py
```diff
@@ -186,6 +187,30 @@ async def test_compute_logprob(self, temperature) -> None:
...
+    async def test_validation(self) -> None:
+        config = {"model": CommonLLMNames.OPENAI_TEST.value, "num_retries": 1}
+        validator = StatefulValidator()
+        llm_op = LLMCallOp(response_validator=validator)
+        await llm_op(config, msgs=[Message(content="Hello")])
+        assert validator.counter == 2  # first attempt should have failed
+
+        config = {"model": CommonLLMNames.OPENAI_TEST.value, "num_retries": 0}
+        llm_op = LLMCallOp(response_validator=StatefulValidator())
+        with pytest.raises(tenacity.RetryError, match="ResponseValidationError"):
```
jamesbraza, line 201: Sort of just sharing, but you can add `reraise=True` to the `tenacity.retry`, and then you can get the original exception as opposed to `tenacity.RetryError`

tests/test_ops.py
```diff
@@ -186,6 +187,30 @@ async def test_compute_logprob(self, temperature) -> None:
...
+    @pytest.mark.asyncio
+    async def test_validation(self) -> None:
+        config = {"model": CommonLLMNames.OPENAI_TEST.value, "num_retries": 1}
+        validator = StatefulValidator()
+        llm_op = LLMCallOp(response_validator=validator)
+        await llm_op(config, msgs=[Message(content="Hello")])
+        assert validator.counter == 2  # first attempt should have failed
+
+        config = {"model": CommonLLMNames.OPENAI_TEST.value, "num_retries": 0}
+        llm_op = LLMCallOp(response_validator=StatefulValidator())
```
jamesbraza, line 200: A simpler way of this is `LLMCallOp(response_validator=MagicMock(side_effect=[ValueError("Invalid response"), None]))`

Basically you can keep things more local to the unit test

ldp/graph/common_ops.py
sidnarayanan: Do you think this should be a warning then? I chose error here to be consistent with this: https://github.com/Future-House/aviary/blob/d46a9537c3522ff554b33870b31212c5c407e4a6/src/aviary/env.py#L245

They're analogous: [agent/env] has a non-fatal error; we report it and continue the rollout.

jamesbraza: Yeah you're right there's a spectrum between warning/error level, so where/how do we draw the line?

I guess we should draw it based on the severity of negative side effect:

- Here: we just re-prompt LLM with same prompt
- Aviary: we add tool failure to history, complete an environment step (which counts toward `max_rollout_steps`), and prompt agent with slightly different message history

I think with Aviary, that side effect is somewhat severe (wasted step, and having the environment possibly change `reward`), whereas in this case the side effect is benign. So maybe I would say the `error` level makes sense for Aviary, but `warning` level makes sense here.

Idk, I am not really passionate here, defer to you and we can just let this one go and change later if it matters.

PR #213, opened 2025-01-18
Fixing `evaluate_consensus`'s `extract_answer_fn` typing

jamesbraza:
When you type it correctly, there's no need for `cast`


ludomitch:
I really dislike casting to satisfy mypy so this makes me very happy

ludomitch (approved):

sidnarayanan (approved):

PR #215, opened 2025-01-22
Fixing `compute_pass_at_k` return type

jamesbraza:
It turns out the return type was not as general as it needed to be

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,6 @@ server = [
     "fastapi>=0.109",  # For Python 3.12 support
 ]
 typing = [
-    "numpy>=2.2.2",  # Pin for np.float typing fixes
     "types-aiofiles",
     "types-tqdm",
 ]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -81,6 +81,7 @@ server = [
     "fastapi>=0.109",  # For Python 3.12 support
 ]
 typing = [
+    "numpy>=2.2.2",  # Pin for np.float typing fixes
```
sidnarayanan, line 84: Can we loosen this and sacrifice typing? IIRC the >=1 pin came from a lack of universal numpy 2 support. I think internal code will break if this is merged. 

I don't remember who/what hit this issue, but we ought to figure it out and add in a comment here.

sidnarayanan (changes requested):

pyproject.toml
jamesbraza: Sure, we don't _need_ this pin, it's just a nice to have, so I reverted the commit

sidnarayanan (approved):

PR #216, opened 2025-01-22
Basic local LLM support

sidnarayanan:
- `LMConfig` captures LLM configuration options
- `TransformerHandler` adds support for buffered LLM execution & multi-device/node distribution

### DIFF
--- a/ldp/nn/handlers/chunking.py
+++ b/ldp/nn/handlers/chunking.py
@@ -1,0 +1,169 @@
+from typing import TypeVar
+
+import torch
+from transformers.generation.utils import GenerateDecoderOnlyOutput
+
+TOutputType = TypeVar("TOutputType", torch.Tensor, GenerateDecoderOnlyOutput)
+
+
+class TensorChunker:
+    """Splits tensors into chunks and adds dummy chunks as needed for parallel processing frameworks like FSDP."""
+
+    def __init__(self, num_chunks, dummy_value=0):
+        self.num_chunks = num_chunks
+        self.dummy_value = dummy_value
+
+    def chunkify(self, *args, **kwargs) -> tuple[list[tuple], list[dict], list[bool]]:
+        """Splits the args into self.num_chunks chunks, adding dummy chunks as needed.
+
+        Returns:
+            tuple: Contains:
+                - A list of len self.num_chunks, where each element is a tuple of split args.
+                - A list of len self.num_chunks, where each element is a dict of split kwargs.
+                - A list of len self.num_chunks, where each element is a bool indicating
+                  whether the chunk is a dummy chunk.
+
+        NOTE: the below Note is not always true - should come back and fix.
+            Can be seen by observing e.g. `len(torch.chunk(torch.arange(20), 8, dim=0)) == 7`.
+            Not necessarily a huge problem - just means we'll be chunking suboptimally.
+            But there may be cases where that causes problems.
+        # Note: Dummy flags will be set only if the incoming batch is smaller than self.num_chunks,
+        # resulting in empty chunks insertion.
+
+        Example:
+        ```python
+        input_tensor = torch.randn(10, 5)  # A tensor with batch size 10
+        some_int = 42  # An integer value that isn't split
+
+        chunker = TensorChunker(num_chunks=4)
+        split_args, split_kwargs, dummy_flags = chunker.chunkify(
+            input_tensor=input_tensor, some_int=some_int
+        )
+        ```
+
+        Output split_args:
+        [
+            ((input_tensor_chunk_0, ), (some_int, )),  # batch of 3
+            ((input_tensor_chunk_1, ), (some_int, )),  # batch of 3
+            ((input_tensor_chunk_2, ), (some_int, )),  # batch of 3
+            ((input_tensor_chunk_3, ), (some_int, )),  # batch of 1
+        ],
+        dummy_flags = [False, False, False, False]
+
+        """
+        # Initialize lists to hold split args and kwargs
+        split_args_list: list[list] = [[] for _ in range(self.num_chunks)]
+        split_kwargs_list: list[dict] = [{} for _ in range(self.num_chunks)]
+        dummy_chunk_flags: list[bool] = []
+
+        # Helper function to process both args and kwargs
+        def process_items(items, is_kwargs=False):
+            nonlocal dummy_chunk_flags
+            for item in items:
+                key, value = item if is_kwargs else (None, item)
+                chunks, out_chunk_dummy_flags = self._split_value(value)
+                for i in range(self.num_chunks):
+                    if is_kwargs:
+                        split_kwargs_list[i][key] = chunks[i]
+                    else:
+                        split_args_list[i].append(chunks[i])
+
+                # Ensure dummy chunk flags are consistent across all items
+                if dummy_chunk_flags and out_chunk_dummy_flags:
+                    assert dummy_chunk_flags == out_chunk_dummy_flags, (
+                        "All inputs must result in the same dummy chunk flags."
+                    )
+                dummy_chunk_flags = out_chunk_dummy_flags or dummy_chunk_flags
+
+        # Process args and kwargs
+        process_items(args)
+        process_items(kwargs.items(), is_kwargs=True)
+
+        split_args_tuples = [tuple(args) for args in split_args_list]
+        return split_args_tuples, split_kwargs_list, dummy_chunk_flags
+
+    @staticmethod
+    def dechunkify(
+        outputs_list: list[TOutputType],
+        dummy_chunk_flags: list[bool],
+    ) -> TOutputType:
+        """Reassembles the outputs from the handlers, removing outputs corresponding to dummy chunks.
+
+        The desired behavior is such that the following are equivalent:
+        ```python
+        # "single device"
+        output_single = module(input)
+
+        # "multi device" (in reality each device would only run one chunk)
+        chunks, dummy_chunk_flags = chunker.chunkify(input)
+        output_multi = chunker.dechunkify(
+            [module(chunk) for chunk in chunks], dummy_chunk_flags
+        )
+        ```
+        """
+        # Filter out outputs corresponding to dummy chunks
+        real_outputs = [
+            output
+            for output, is_dummy in zip(outputs_list, dummy_chunk_flags, strict=True)
+            if not is_dummy
+        ]
+
+        if isinstance(real_outputs[0], torch.Tensor):
+            # If we received a list of tensors, concat along the batch dimension
+            # TODO: consider how/if to handle differently shaped tensors
+            return torch.cat(real_outputs, dim=0)
+
+        if isinstance(real_outputs[0], GenerateDecoderOnlyOutput):
+            sequences = []
+            for output in real_outputs:
+                sequences.extend(output.sequences)
+            scores: list[list[torch.Tensor | None]] | None = None
+            if real_outputs[0].scores is not None:
+                assert all(output.scores is not None for output in real_outputs)
+                # `output.scores` is a tuple (one el per decoded token), with each tensor of shape
+                # (batch_size, vocab_size).
+                # To get the scores for the `i`-th batch element, we'd do `[score[i] for score in output.scores]`.
+                # `output.scores` will have different lengths across workers, so we cannot simply concatenate them.
+                # Adding dummy scores is error-prone and loses info. Instead, we create a merged `scores` with the
+                # same access semantics. `score[i]` will be a tensor if it exists for batch element `i` and `None`
+                # otherwise.
+                scores = []
+                max_output_len = max(len(output.scores) for output in real_outputs)
+                for i in range(max_output_len):
+                    scores_step_i: list[torch.Tensor | None] = []
+                    for output in real_outputs:
+                        if i < len(output.scores):
+                            scores_step_i.extend(list(output.scores[i]))
+                        else:
+                            # Add bsz Nones for this worker, since it did not reach token `i`.
+                            scores_step_i.extend([None] * output.sequences.shape[0])
+
+                    scores.append(scores_step_i)
+
+            return GenerateDecoderOnlyOutput(
+                sequences=sequences,
+                scores=scores,
+            )
+
+        raise ValueError("Unsupported output type")
+
+    def _split_value(self, value):
+        """
+        Splits a single value into chunks, adding a dummy value if necessary.
+
+        Right now, only torch.Tensor values are split. Non-tensor values are replicated.
+        """
+        if isinstance(value, torch.Tensor):
+            chunks = list(torch.chunk(value, self.num_chunks, dim=0))
+            dummy_chunk_flags = []
+            for i in range(self.num_chunks):
+                if i >= len(chunks):
+                    # Chunk 0 will always exist, and we need only a batch of one to activate the model
+                    chunks.append(torch.full_like(chunks[0][:1], self.dummy_value))
+                    dummy_chunk_flags.append(True)
+                else:
+                    dummy_chunk_flags.append(False)
+
+            return chunks, dummy_chunk_flags
+        # Non-tensor values are replicated
+        return [value] * self.num_chunks, None
--- a/ldp/nn/handlers/transformer_handler.py
+++ b/ldp/nn/handlers/transformer_handler.py
@@ -1,0 +1,1017 @@
+from __future__ import annotations
+
+import logging
+import os
+import socket
+import sys
+from abc import ABC, abstractmethod
+from collections.abc import Awaitable, Callable
+from enum import StrEnum, auto
+from functools import cache, partial, wraps
+from pathlib import Path
+from typing import Any, Concatenate, ParamSpec, Self, TypeVar, cast
+
+import accelerate
+import torch
+import torch.distributed as dist
+import tree
+from dask import config
+from dask.distributed import Client
+from dask_jobqueue import SLURMCluster
+from pydantic import BaseModel, ConfigDict, Field, field_validator
+from torch import nn
+from torch.cuda import nccl
+from torch.distributed.fsdp import (
+    BackwardPrefetch,
+    FullStateDictConfig,
+    FullyShardedDataParallel,
+    MixedPrecision,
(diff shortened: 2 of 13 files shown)


ldp/nn/handlers/chunking.py
```diff
@@ -0,0 +1,169 @@
...
+
+        Right now, only torch.Tensor values are split. Non-tensor values are replicated.
+        """
+        if isinstance(value, torch.Tensor):
+            chunks = list(torch.chunk(value, self.num_chunks, dim=0))
+            dummy_chunk_flags = []
+            for i in range(self.num_chunks):
+                if i >= len(chunks):
+                    # Chunk 0 will always exist, and we need only a batch of one to activate the model
+                    chunks.append(torch.full_like(chunks[0][:1], self.dummy_value))
```
jamesbraza, line 162: Can you document the `[:1]`?

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1017 @@
...
+    from typing import overload
+else:
+    from typing_extensions import overload  # noqa: UP035
+
+try:
+    from dask_cuda import LocalCUDACluster
+except ImportError:
+    LocalCUDACluster = None
+
+config.set({
```
jamesbraza, line 54: Given we are a package people are installing, shouldn't we not being doing module-level configuration? Can we move this to a configuration function that one can import/use?

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1017 @@
...
+})
+
+logger = logging.getLogger(__name__)
+
+TReturn = TypeVar("TReturn")
+TParams = ParamSpec("TParams")
+
+
+def is_conversation(messages) -> bool:
+    """Check if messages is an instance of Conversation."""
```
jamesbraza, line 73: What is a Conversation? Can you clarify where it comes from?

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1017 @@
...
+        # If model is not FSDP, this context manager is a no-op.
+        with FullyShardedDataParallel.summon_full_params(model, recurse=False):
+            logger.debug(
+                f"model.generate() input_ids shape: {kwargs['input_ids'].shape}, rank {os.environ.get('RANK')}"
+            )
+            return model.generate(
+                *args,
+                **kwargs,
+                pad_token_id=model.config.pad_token_id,  # not always set properly by .generate()
+                # eos_token_id=model.config.eos_token_id,  TODO remove
```
jamesbraza, line 210: Can we just remove this now?

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1017 @@
...
+        os.environ["WORLD_SIZE"] = str(self.world_size)
+        os.environ["LOCAL_RANK"] = str(self.local_rank)
+        os.environ["MASTER_ADDR"] = self.master_addr
+        os.environ["MASTER_PORT"] = str(self.master_port)
+
+        # These tell PretrainedModel.from_pretrained() that we're using FSDP
+        os.environ["ACCELERATE_USE_FSDP"] = "1"
+        os.environ["FSDP_CPU_RAM_EFFICIENT_LOADING"] = str(
+            int(self.cpu_ram_efficient_loading)
+        )
```
jamesbraza, line 338: Feel free to ignore this, but for huge env var updates, I like bulk updates `os.environ.update({...})` as opposed to individual sets

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1017 @@
...
+
+        if self.config.lm_config.device == "meta":
+            self.module = prepare_model_for_fsdp_with_meta_device(self.module)
+
+        self.module = self.accelerator.prepare(self.module)
+
+    def set_seed(self, seed: int) -> None:
+        """Set the seed for the current worker."""
+        set_seed(seed)
```
jamesbraza, line 396: Do we need this method? Can we just have the caller import/use `set_seed` themselves?

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1017 @@
...
+                threads_per_worker=parallel_mode_config.num_cpus_per_worker,
+                host=parallel_mode_config.scheduler_addr,
+                port=parallel_mode_config.scheduler_port,
+                memory_limit=None,  # do not let Dask manage memory - if we OOM, we OOM
+            )
+        except TypeError:
+            raise ImportError(
+                f"To run {type(self).__name__} with execution mode"
+                f" {parallel_mode_config.execution_mode}, please install the"
+                " 'dask-cuda' package by running `make nn` in the repo root."
```
jamesbraza, line 483: `make nn` is not a thing here, can you update this?

ldp/nn/handlers/chunking.py
kwanUm: # Chunk 0 will always exist, and we need only a batch of one to activate the model

Maybe add to that existing comment above "therefore [:1] is used"

ldp/nn/handlers/transformer_handler.py
sidnarayanan: I'll push this to another PR

ldp/nn/handlers/transformer_handler.py
sidnarayanan: I've removed the try/except altogether.

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1017 @@
...
+# From: https://github.com/pytorch/torchtune/blob/c5db813ce0473db090a4f1f6b450f559acac58e5/torchtune/training/_distributed.py#L207
+# NOTE: I removed the LoRA logic from the source code since we don't use torchtune for now.
+# See issue in docstring for why we needed this.
+def prepare_model_for_fsdp_with_meta_device(model: nn.Module) -> nn.Module:
+    """Dynamically define reset_parameters on every submodule of the model.
+
+    More details here: https://github.com/pytorch/pytorch/issues/104187.
+
+    Args:
+        model: model class to prepare for usage with FSDP and meta device.
```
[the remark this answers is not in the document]
sidnarayanan: Removed "class"

ldp/nn/handlers/transformer_handler.py
sidnarayanan: Yeah I see what you mean, but I'll think about this later. We're doing lots of env-level configuration, like setting CUDA default settings

ldp/nn/handlers/transformer_handler.py
sidnarayanan: Oops, this will make more sense with an upcoming PR, which will introduce `Conversation`

ldp/nn/handlers/chunking.py
```diff
@@ -0,0 +1,170 @@
...
+        dummy_flags = [False, False, False, False]
+
+        """
+        # Initialize lists to hold split args and kwargs
+        split_args_list: list[list] = [[] for _ in range(self.num_chunks)]
+        split_kwargs_list: list[dict] = [{} for _ in range(self.num_chunks)]
+        dummy_chunk_flags: list[bool] = []
+
+        # Helper function to process both args and kwargs
+        def process_items(items, is_kwargs=False):
```
jamesbraza, line 60:
```suggestion
        def process_items(items, is_kwargs: bool = False) -> None:
            """Helper to add args/kwargs into above lists."""
```

This comment should be a docstring

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    offload_cpu: bool = Field(
+        default=False,
+        description="Whether to offload model parameters and gradients to CPU.",
+    )
+    activation_checkpointing: bool = False
+    cpu_ram_efficient_loading: bool = False
+    state_dict_type: StateDictType = StateDictType.FULL_STATE_DICT
+    backward_prefetch: BackwardPrefetch | None = Field(
+        BackwardPrefetch.BACKWARD_PRE,
```
jamesbraza, line 104: Do we need to allow `| None` here if we have a default? And if we want `| None`, can you document what that means in the `description`

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1007 @@
...
+
+class FSDPConfig(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    offload_cpu: bool = Field(
+        default=False,
+        description="Whether to offload model parameters and gradients to CPU.",
+    )
+    activation_checkpointing: bool = False
+    cpu_ram_efficient_loading: bool = False
```
jamesbraza, line 101: Do you mind adding a `description` to this one, stating what it integrates with? Also, feel free to punt on this one

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1007 @@
...
+        description=(
+            "Execution mode of the current setup, defines how to allocate resources"
+            " (cpus/gpus) for the model to run on."
+        ),
+        default=ExecutionMode.LOCAL_MACHINE,
+    )
+
+    scheduler_addr: str = "localhost"
+    scheduler_port: int = Field(default=0, description="0 means Dask picks randomly.")
+    torch_port: int = Field(default_factory=get_unused_port)
```
jamesbraza, line 137: Can you add a `description` to what `torch_port` means? To me, `torch` is a framework, not an OS.

Maybe just do `ParallelWorkerConfig.model_fields["xyz"].description` to make the linkage clear

So maybe we can make the name more verbose, or add a description stating the implications

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    # Below configurations (walltime, memory, log_directory) are only for ExecutionMode.SLURM_CLUSTER:
+    walltime: str = Field(
+        default="00:30:00", description="Max time the worker can run."
+    )
+    memory: str = Field(default="32GB", description="Memory allocated per worker.")
+    log_directory: str = Field(
+        default=f"{REPO_ROOT}/logs/slurm_outputs/",
```
jamesbraza, line 145: Any chance we can allow `str | os.PathLike`?

Also, this `default`, it's not a Windows compliant path. Can you use `pathlib.Path` or `os.path` to make the default?

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1007 @@
...
+            case LMType.GENERATION:
+                tokenizer, model = config.lm_config.get_causal_lm()
+                # On left for https://github.com/huggingface/transformers/pull/7552
+                # ^ that seems to work for most HF models w/ absolute position embeddings
+                # Left padding always works for relative position embeddings
+                tokenizer.padding_side = "left"
+            case LMType.REGRESSION:
+                tokenizer, model = config.lm_config.get_regression_lm()
+            case _:
+                raise NotImplementedError(f"Unsupported LM type: {config.lm_type}")
```
jamesbraza, line 230:
```suggestion
                assert_never(config.lm_type)
```

Imo this `NotImplementedError` is a bit misleading because it's not that we don't support it, but it's because the `LMType` enum has no more values. So we can rely on type checker to eliminate this branch for us

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1007 @@
...
+        logger.info(f'Loading checkpoint from "{ckpt}"')
+        self.accelerator.load_state(str(ckpt), **kwargs)
+
+    def save_checkpoint(self, ckpt: os.PathLike | str, **kwargs) -> None:
+        self.accelerator.save_state(str(ckpt), **kwargs)
+        self.barrier()
+        # We do not want to save random states - they would be loaded by load_state
+        # automatically. Clean up after all processes have saved.
+        if int(os.getenv("RANK", "0")) == 0:
+            random_states = Path(ckpt).rglob("random_state*pkl")
```
jamesbraza, line 264:
```suggestion
            random_states = Path(ckpt).rglob("random_state*.pkl")
```

Just capturing extension

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1007 @@
...
+            for random_state in random_states:
+                random_state.unlink()
+
+    @staticmethod
+    def barrier() -> None:
+        if dist.is_initialized():
+            dist.barrier()
```
jamesbraza, line 271: [style nit]

This should be named like `make_barrier` or `create_barrier`, methods should be named as if some action

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1007 @@
...
+
+        match parallel_mode_config.execution_mode:
+            case ExecutionMode.LOCAL_MACHINE:
+                self._init_local_cluster(config, parallel_mode_config)
+            case ExecutionMode.SLURM_CLUSTER:
+                self._init_slurm_cluster(config, parallel_mode_config)
+            case _:
+                raise NotImplementedError(
+                    f"Unsupported execution mode: {parallel_mode_config.execution_mode}"
+                )
```
jamesbraza, line 448:
```suggestion
                typing.assert_never(parallel_mode_config.execution_mode)
```

Same as above, this `NotImplementedError` is misleading because actually, it's an impossible branch, not an unimplemented branch

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1007 @@
...
+            processes=1,  # Single dask worker per slurm worker
+            walltime=parallel_mode_config.walltime,
+            job_extra=[
+                "--gres=gpu:1"
+            ],  # 1 GPU per worker seems to be the common case for now
+            log_directory=parallel_mode_config.log_directory,
+        )
+        self._initialize_workers(config, parallel_mode_config)
+
+    def _initialize_workers(self, config, parallel_mode_config):
```
jamesbraza, line 491:
```suggestion
    def _initialize_workers(self, dask_client_config, parallel_mode_config):
```

Having this take a `config` and a `parallel_mode_config` makes one wonder, what is this first config? Let's be more verbose on the name for the first `config`

ldp/nn/handlers/transformer_handler.py
```diff
@@ -0,0 +1,1007 @@
...
+            # Warn the user about potential training issues
+            logger.warning(
+                "Tokenizer's chat_template does not contain '{% generation %}'. "
+                "Training will have issues as the tag does not exist, which means "
+                "HuggingFace's internal code for retrieving the assistant_mask for training will be invalid."
+            )
+            return
+
+    with open(template_path) as file:  # noqa: FURB101
+        loaded_chat_template = file.read()
```
jamesbraza, line 1002: Jw, why not listen to `FURB101`? I like this rule because it's less code

ldp/nn/lm_config.py
```diff
@@ -0,0 +1,181 @@
...
+class ModelMode(StrEnum):
+    training = "training"
+    inference = "inference"
+
+
+class TorchDType(StrEnum):
+    bf16 = "bfloat16"
+    fp16 = "float16"
+    fp32 = "float32"
+    auto = "auto"
```
jamesbraza, line 29:
```suggestion
    unspecified = "unspecified"
```

It seems like `auto` is not really the right word here, it's more like `unspecified`.

And if I am wrong, can you add an inline code comment stating what `auto` means here?

pyproject.toml
```diff
@@ -71,7 +71,14 @@ monitor = [
...
 ]
 nn = [
-    "torch>=2.2",
+    "accelerate",
+    "dask-cuda>=24.8.2",
+    "dask-jobqueue",
+    "dask[distributed]",
+    "tokenizers>0.20",
+    "torch>=2.5",
+    "transformers>=4.46",
```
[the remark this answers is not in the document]
sidnarayanan: Honestly I'm just pinning to minimum versions that I know work - it may be possible to loosen

ldp/nn/handlers/transformer_handler.py
sidnarayanan: Yeah None is a valid option to pass to FSDP (i.e. no prefetch). The link I put in the description describes this.

ldp/nn/handlers/transformer_handler.py
sidnarayanan: I don't want to overload our code with docstrings - it makes the code less readable. These are all FSDP arguments that are well-documented in the FSDP docs.

ldp/nn/handlers/transformer_handler.py
sidnarayanan: I don't understand the framework/OS comment, nor the `model_fields` suggestion.

In any case, I think `torch_port` is clear - it's the port that torch uses to communicate between processes. Fairly standard configuration that should be familiar to users who understand torch distributed.

ldp/nn/handlers/transformer_handler.py
sidnarayanan: I'll circle back to this. I don't want the logs to live inside the repo long-term anyway.

ldp/nn/handlers/transformer_handler.py
sidnarayanan: I disagree - I am aliasing `torch.distributed.barrier()`.

ldp/nn/handlers/transformer_handler.py
sidnarayanan: It's not a `dask_client_config` though. I have added typing to avoid ambiguity.

ldp/nn/handlers/transformer_handler.py
sidnarayanan: Hm I don't remember. Switching to `template_path.open()`.

ldp/nn/lm_config.py
sidnarayanan: It's already documented here: https://huggingface.co/docs/transformers/v4.48.0/en/main_classes/model#transformers.PreTrainedModel.from_pretrained.torch_dtype . Per above, I don't want to re-document parameters we are forwarding

PR #217, opened 2025-01-22
Local LLM Op and Agent

sidnarayanan:
- `LocalLLMCallOp` is an Op that samples from a pytorch LLM
- `SimpleLocalAgent` is like `SimpleAgent`, but uses the above
- Chat templates are made configurable, with a few more commonly-used ones added to the repo

PR #219, opened 2025-01-23
Local LLM call arguments

sidnarayanan:
Missed a couple things in #217 :
- Ability to pass in arbitrary call kwargs
- Support for a custom logit processor base class that can be finalized post-generation. The implementations will come in a subsequent PR.

### DIFF
--- a/ldp/nn/agent/simple_local_agent.py
+++ b/ldp/nn/agent/simple_local_agent.py
@@ -45,9 +45,7 @@ class AgentLMConfig(_LMConfig):
     @field_validator("llm_call_kwargs")
     @classmethod
     def validate_llm_call_kwargs(cls, v: dict) -> dict:
-        v.setdefault("top_k", None)
-        v.setdefault("top_p", 1.0)
-        return v
+        return {"top_k": None, "top_p": 1.0} | v
 
 
 # TODO: consider merging with SimpleAgent
--- a/ldp/nn/agent/simple_local_agent.py
+++ b/ldp/nn/agent/simple_local_agent.py
@@ -3,6 +3,7 @@ from typing import cast
 import torch
 import torch.distributed as dist
 from aviary.core import Message, Tool, ToolRequestMessage
+from pydantic import Field, field_validator
 
 from ldp.agent import Agent, SimpleAgentState
 from ldp.graph import OpResult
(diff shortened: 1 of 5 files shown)


ldp/nn/agent/simple_local_agent.py
```diff
@@ -33,6 +34,21 @@ class AgentLMConfig(_LMConfig):
...
+        "are better defaults than HF's.",
+        validate_default=True,
+    )
+
+    @field_validator("llm_call_kwargs")
+    @classmethod
+    def validate_llm_call_kwargs(cls, v: dict) -> dict:
+        v.setdefault("top_k", None)
+        v.setdefault("top_p", 1.0)
+        return v
```
jamesbraza, line 50:
```suggestion
        return {"top_k": None, "top_p": 1.0} | v
```

Note with this behavior, deserialization gets affected. We have been burned by this type of stuff before

Can we not in-place edit the input `v`, instead doing so via something like this

ldp/nn/generation/base.py
```diff
@@ -0,0 +1,10 @@
+from abc import ABC, abstractmethod
+
+import torch
+from transformers import LogitsProcessor
+
+
+class LogitsProcessorWithFinalize(LogitsProcessor, ABC):
+    @abstractmethod
+    def finalize(self, token_ids: torch.Tensor) -> None:
+        pass
```
jamesbraza, line 10: Can you add a docstring here? It's customary in an `abstractmethod` to detail what subclassers should do.

Imo the signature alone isn't very intuitive, I think it needs a docstring

ldp/nn/agent/simple_local_agent.py
```diff
@@ -33,6 +34,19 @@ class AgentLMConfig(_LMConfig):
...
+        default_factory=dict,
+        description="Additional kwargs to pass to LocalLLMCallOp.forward. "
+        "Note that the validator defaults top_k=None and top_p=1.0, which "
+        "are better defaults than HF's.",
+        validate_default=True,
+    )
+
+    @field_validator("llm_call_kwargs")
+    @classmethod
+    def validate_llm_call_kwargs(cls, v: dict) -> dict:
```
jamesbraza, line 47:
```suggestion
    def inject_preferred_llm_call_defaults(cls, v: dict) -> dict:
```

I might name this closer to its purpose. It's not really a validator

jamesbraza (approved):

PR #220, opened 2025-01-23
Make dask_cuda import lazy

sidnarayanan:
`dask_cuda` only works on Linux


ldp/nn/handlers/transformer_handler.py
```diff
@@ -483,6 +482,9 @@ def _init_local_cluster(
         self, config: TransformerHandlerConfig, parallel_mode_config: ParallelModeConfig
     ):
         """Initialize a Dask cluster on local machine."""
+        # lazy import since LocalCUDACluster only works on Linux machines
```
jamesbraza, line 485:
```suggestion
        # lazy import since dask-cuda only works on Linux machines
```

jamesbraza (approved):

PR #222, opened 2025-01-24
Created `ClearOptimizerBuffersCallback`

jamesbraza:
See PR title

### DIFF
--- a/ldp/alg/optimizer/opt.py
+++ b/ldp/alg/optimizer/opt.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
 import logging
 from abc import ABC, abstractmethod
 from collections.abc import Iterable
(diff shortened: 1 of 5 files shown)


ldp/alg/optimizer/replay_buffers.py
```diff
@@ -62,6 +62,14 @@ def batched_iter(
     ) -> Iterator[dict]:
         return self._batched_iter(self.data, batch_size, shuffle, infinite)
 
+    async def reset(self) -> None:
```
sidnarayanan, line 65: What do you think about naming this `clear()` That's kind of the idiomatic python term for emptying an object

sidnarayanan (approved):

ldp/alg/optimizer/replay_buffers.py
jamesbraza: Sure, done. Turns out `UserList` already has a `clear` method for us ✅

ldp/alg/optimizer/opt.py
```diff
@@ -60,3 +62,14 @@ def aggregate(
     async def update(self) -> None:
         for optimizer in self.optimizers:
             await optimizer.update()
+
+
+class OptimizerClearBufferCallback(Callback):
```
sidnarayanan, line 67: Can you rename this to `ClearOptimizerBuffersCallback`? The intent is clearer

sidnarayanan: Also can you move this to `ldp.alg.callbacks`?

jamesbraza: Both are now done, thank you!

PR #223, opened 2025-01-25
Configurable message parsing

sidnarayanan:
`LocalLLMCallOp` previously hardcoded llama 3.1-friendly message parsing logic. This PR makes that configurable by the user.

### DIFF
--- a/ldp/nn/graph/llm_call_op.py
+++ b/ldp/nn/graph/llm_call_op.py
@@ -1,12 +1,19 @@
 from __future__ import annotations
 
 import json
+from collections.abc import Callable
 from functools import partial
 from typing import Any, ClassVar
 
 import tree
-from aviary.core import MalformedMessageError, Message, Messages
-from aviary.tools import Tool, ToolCall, ToolRequestMessage
+from aviary.core import (
+    MalformedMessageError,
+    Message,
+    Messages,
+    ToolCall,
+    ToolRequestMessage,
+    Tools,
+)
 from transformers import LogitsProcessorList
 
 from ldp.graph.gradient_estimators import assign_constant_grads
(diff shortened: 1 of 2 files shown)


ldp/nn/graph/llm_call_op.py
```diff
@@ -23,71 +31,76 @@
...
+
+class Llama31Parser(MessageAndToolParser):
+    """Follows the Llama 3.1 syntax.
+
+    See details:
+    https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_1/#-tool-calling-(8b/70b/405b)-
+    """
+
+    supported_templates: ClassVar[set[str]] = {
+        "llama2_chat_template_ori.jinja",
```
jamesbraza, line 78: No one but us is going to know what an Ori template is. Any chance you can rename this file, or add a comment explaining it?

ldp/nn/graph/llm_call_op.py
```diff
@@ -23,71 +31,76 @@
...
+        "llama3.1_chat_template_vllm.jinja",
+        "llama3_chat_template_ori.jinja",
+    }
+
+    @classmethod
+    def get_message_content(cls, msg: Message) -> str | None:
+        if isinstance(msg, ToolRequestMessage):
+            assert len(msg.tool_calls) == 1, (
+                "Support parsing only single tool call for now"
+            )
```
jamesbraza, line 91: Since this is about supporting something, I would rather see `raise NotImplementedError` than `assert`

ldp/nn/graph/llm_call_op.py
```diff
@@ -23,71 +31,76 @@
...
-            collate_fn=partial(
-                collate_fn_transformer_left_pad, pad_token_id=pad_token_id
-            ),
-            decollate_fn=decollate_fn_transformer_decoder,
-        )
-        self.model_handler = handler_config.make_async_module()
-        self.model_name = model_config.model
+    @classmethod
+    @abstractmethod
+    def prep_tools_for_tokenizer(cls, tools: Tools | None) -> list[dict] | None:
```
jamesbraza, line 49: If we are going to prepare tools, why are we then passing `None` tools?

Just pointing out, `Tools | None` doesn't really make sense, imo this should be something like `*tools: Tool` or `tools: Tools`

jamesbraza (approved):
LGTM, added some optional comments

ldp/nn/graph/llm_call_op.py
sidnarayanan: It's because `tools` can be `None` here: https://github.com/Future-House/ldp/blob/0d0e0bd27ad6e91aa271aec5c72b7ee31628a12a/ldp/graph/common_ops.py#L256 - this method just mirrors the typing of the arguments that can be passed to `[Local]LLMCallOp`.

ldp/nn/graph/llm_call_op.py
sidnarayanan: Yeah it's on my to-do list to rename this, and there's documentation in `chat_templates/README.md` for all of the chat templates. I'll leave it for a later PR to sort this out.

PR #226, opened 2025-01-28
Propagates llmclient changes to ldp

maykcaldas:
This is just a draft. A new `fh-llm-client` release is still needed.

TODO:
- [x] Merge https://github.com/Future-House/llm-client/pull/48 to fix ldp issues
- [x] Update paper-qa to new llmclient and ldp
- [ ] Release `llmclient` `v0.1.0`
- [ ] Update `fh-llm-client` dependency in `pyproject.toml`
- [x] Go through cassettes and check if all changes are needed
- [ ] Re-test this PR
- [ ] Merge


ldp/graph/common_ops.py
```diff
@@ -270,24 +270,27 @@ async def forward(
...
+        model = LLMModel(name=config["name"], config=config)
 
         if not tools:
             # if no tools are provided, tool_choice must be 'none'
             tool_choice = "none"
 
         result = await self._call_single_and_maybe_validate(
             model=model,
-            num_retries=config.get("num_retries", 0),
+            # num_retries=config.get("num_retries", 0),
```
jamesbraza, line 281: Why comment this out?

ldp/graph/common_ops.py
```diff
@@ -311,7 +314,7 @@ async def forward(
         return result.messages[0]
 
     async def _call_single_and_maybe_validate(
-        self, model: LLMModel, num_retries: int, **kwargs
+        self, model: LLMModel, num_retries: int = 0, **kwargs
```
jamesbraza, line 317: Why are we adding a default here now?

ldp/graph/common_ops.py
```diff
@@ -270,24 +270,27 @@ async def forward(
...
             tool_choice=tool_choice,
+            **config,
         )
         if result.messages is None:
             raise ValueError("No messages returned")
 
         # if not set, assume temp = 1. TODO: when would it not be set?
-        temperature: float = (result.config or {}).get("temperature", 1.0)
+        # ASK: Should LLMResult carry config? Currently LiteLLMModel keeps it.
+        # temperature: float = (result.config or {}).get("temperature", 1.0)
```
jamesbraza, line 292: Are you saying that we can generalize `LLMResult` to have a link to the entire input configuration, not just `temperature`?

tests/test_agents.py
```diff
@@ -89,7 +89,7 @@ def many_edge_cases(
 
 
 class TestAgentState:
-    @pytest.mark.vcr
+    @pytest.mark.vcr(match_on=[*VCR_DEFAULT_MATCH_ON, "body"])
```
jamesbraza, line 92: Why do we need this now? Let's try to avoid extra changes if possible

If we do need it, do you mind adding a comment why we need this?

I guess, let's avoid extra changes in this PR

ldp/graph/common_ops.py
```diff
@@ -366,8 +369,9 @@ async def compute_logprob(
         # TODO: possibly move to MultipleCompletionLLMModel here, though we need to check that the estimates
         # are consistent - not sure we'd be sampling from the same distribution as N independent samples.
         # TODO: think about whether sampling params besides temperature need to be accounted for, like top_p
+        # NOTE: messages is being passed as a model_kwargs. This is raising the [call-arg] error.
         results = await asyncio.gather(*[
-            model.call_single(temperature=1, **model_kwargs)
+            model.call_single(temperature=1, **model_kwargs)  # type: ignore[call-arg]
```
jamesbraza, line 374: Are you sure this is necessary? Running `mypy` locally it doesn't seem necessary

ldp/graph/common_ops.py
maykcaldas: Forgot to push the comments lol

I am passing **config as kwargs so that they can be passed to `call_single`. It created a conflict because num_retries was being passed twice in some cases. Then I moved the default

ldp/graph/common_ops.py
maykcaldas: Check previous comment

ldp/graph/common_ops.py
maykcaldas: Removed it. I am not sure why, locally, mypy complained about temperature not being an attribute of call_single. Probably an outdated pre-commit environment

ldp/graph/common_ops.py
maykcaldas: Originally, it was getting temperature from `result.config` (the code commented out). Which is weird, because I couldn't see`LLMResult.config` being populated anywhere. Temperature is an attribute from `LiteLLMModel`. 
Then I'm getting `temperature` from the `config` passed to `LLMCallOp.forward` and defaulting it to `1.0`  as it was before. 

We could generalize LLMResult to have a "frame" of the configuration used when the result was generated. But I don't think it's important for this PR

jamesbraza: Okay gotchu. I am not sure. I would say whatever you do, just make sure we have test coverage with assertions

ldp/graph/common_ops.py
jamesbraza: One way to do this is something like `{"num_retries": config.get("num_retries", 0)} | kwargs`

However, I like what you have done, I think it makes sense to default `num_retries` to 0. Resolving this one

ldp/graph/common_ops.py
jamesbraza: Resolving this one

ldp/graph/common_ops.py
maykcaldas: Cool. This is being tested in `test_compute_logprob`

tests/test_agents.py
maykcaldas: Think we solved it! 🤘🏻

jamesbraza (commented):
Do you mind merging with `main`, and do we need all these `cassette` changes?

tests/test_ops.py
```diff
@@ -164,18 +164,24 @@ async def generate_story() -> str:
...
         llm_call_op = LLMCallOp()
         message_result = await llm_call_op(
-            LLMModel.model_fields["config"].default_factory(),  # type: ignore[call-arg, misc]
+            config=LLMModel.model_fields["config"].default_factory(),  # type: ignore[call-arg, misc]
             msgs=[Message(content="Hello")],
             tools=[],
         )
-        assert isinstance(message_result.value, ToolRequestMessage)
-        assert not message_result.value.tool_calls
+        # ASK: Should llm_call_op return a ToolRequestMessage once there's no tool calls?
```
jamesbraza, line 171: I think it's fine to have a `ToolRequestMessage` without tool calls. What makes you feel like we need to change this behavior?

I think when we pass `tool`s, even if empty, it signifies `ToolRequestMessage` will be returned. We have typing support around this as well in `LLMCallOp.forward` I think

maykcaldas: I'm not sure what was the behavior before. If we don't pass tools to the LLM call, we set `tool_choice` to `none` here: https://github.com/Future-House/ldp/blob/e2d7df8f6e3706695530472f48dad724d1b959c7/ldp/graph/common_ops.py#L275-L285

Which means that [no tool will be called](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice), according to OpenAI doc.

Then, in `llmclient`, we check if the generate message in the completion is a `tool_calls` and create the Message accordingly: https://github.com/Future-House/llm-client/blob/3831881f72bc3e726d9d856ea3fa764a015b2f58/llmclient/llms.py#L576-L588

But it won't create a `ToolRequestMessage` because a call with an empty `tools` list will never return a `tool_calls`. 🤔

jamesbraza: I guess we should think of our test suite as a specification for the code's behavior. I don't think the LLM client refactor should come with necessitating changes in the downstream LDP.

Take a look here: https://github.com/Future-House/ldp/blob/v0.22.0/ldp/graph/common_ops.py#L234-L250
We see the `overload` defining the presence of `tools`, even if it's an empty list, as making a `ToolRequestMessage`.

Can we preserve that behavior?

tests/test_ops.py
```diff
@@ -159,12 +159,12 @@ async def generate_story() -> str:
         # Environment tracks its internal costs
         assert env.total_cost > 0
 
-    @pytest.mark.vcr
+    # @pytest.mark.vcr
```
jamesbraza, line 162: Why comment this? Can you uncomment this?

Let's try to keep the diff minimal

tests/test_ops.py
```diff
@@ -159,12 +159,12 @@ async def generate_story() -> str:
...
     @pytest.mark.asyncio
     async def test_empty_tools(self) -> None:
         llm_call_op = LLMCallOp()
         message_result = await llm_call_op(
-            LLMModel.model_fields["config"].default_factory(),  # type: ignore[call-arg, misc]
+            config=LLMModel.model_fields["config"].default_factory(),  # type: ignore[call-arg, misc]
```
jamesbraza, line 167: Can we not use kwarg here? I think test is better with positional arg, and this bloats the diff

maykcaldas: I understand we want to keep the changed minimal and that this bloats the diff. So I removed it.

But why do you think positional args are better in test? To catch changes in the arguments list/new arguments?

jamesbraza: So maybe to compare, we can have this:

```python
some_tuple: tuple

# Extracts values, but doesn't also confirm length
value_1, value_2 = some_tuple[0], some_tuple[1]

# Extracts values and confirm length
value_1, value_2 = some_tuple
```

We have two implementations with the same result of extracting values, but the second is tighter than the first.

It's a similar story here, but not quite the same.

Basically this change from positional to keyword changes the semantics of this test. Since we have places in our code base that use the positional arg, it's nice to protect that within our test suite.

Does that make sense?

Also, it's nice to avoid diff churn

jamesbraza (approved):
LGTM, nice work here @maykcaldas

jamesbraza (approved):

PR #231, opened 2025-02-18
Rollout Manager & Chunking Improvements

kwanUm:
1) Adds a progress bar exception counter in `rollout.py` to improve visibility of errors - with configurable immediate or summary logging of errors
2) Enhances chunking behavior by using real data in dummy chunks, reducing model errors.  
3) Updates parallel LLM operations to use NVML for better GPU detection when running Torch FSDP from spawned processes. 
4) Improves error handling in Dask workers by replacing `client.gather()` with `dask.wait()`, waiting on first exception, and reporting exception in worker immediately as they come. Assuming here that dasks exception should crash the program.
5) Simplified sample_trajectories() API with optional parameters


ldp/nn/handlers/transformer_handler.py
```diff
@@ -463,6 +472,8 @@ def __init__(self, config: TransformerHandlerConfig):
 
         self._initialized = True
 
+        atexit.register(self.teardown)
```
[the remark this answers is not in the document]
kwanUm: Wouldn't that close only the dask cluster of self rather than destroy all dask clusters?

src/ldp/nn/agent/simple_local_agent.py
```diff
@@ -109,6 +115,20 @@ async def get_asv(
 
         # Update state messages with result and return the new state
         next_state.messages = [*next_state.messages, result.value]
+        
+          
+        import ipdb; ipdb.set_trace()
```
jamesbraza, line 120: Can you remove `ipdb`

src/ldp/alg/rollout.py
```diff
@@ -193,21 +196,52 @@ async def _sample_trajectories_from_envs(
...
+                        exception_counter[exc_str] += 1
+                        num_exceptions = sum(exception_counter.values())
+                        pbar.set_postfix({"num_exceptions": num_exceptions})
+
+        # Final summary of exceptions (if any)
+        if exception_counter:
+            logger.info("Caught exceptions:")
+            logger.info("%-6s %-50s", "Count", "Exception")
+            for exc, count in exception_counter.items():
+                logger.info("%-6d %-50s", count, exc)
```
jamesbraza, line 235: Can you build a string here and do one `logger.info`?

With any concurrency otherwise in log aggregators, these will get broken up

src/ldp/nn/agent/simple_local_agent.py
```diff
@@ -42,6 +46,10 @@ class AgentLMConfig(_LMConfig):
         ),
         validate_default=True,
     )
+    max_traj_token_count: int | None = Field(
+        default=None,
+        description="If set, raise an error if the total tokens in the trajectory exceed this value.",
+    )
```
sidnarayanan, line 52: Can we call this `max_messages_token_count` or something? The messages passed in are not necessarily the full trajectory (i.e. if we are collapsing previous observations)

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -192,6 +194,14 @@ async def __call__(  # type: ignore[override]
     @staticmethod
     def model_generate(model: PreTrainedModel, *args, **kwargs):
         """A method that can be used as module_call_fn to sample from an LLM."""
+        if int(os.environ.get("WORLD_SIZE", "1")) > 1:
+            synced_gpus = kwargs.pop("synced_gpus", None)
+            if synced_gpus is None:
+                logger.debug("synced_gpus not defined, defaulting to True.")
+                kwargs["synced_gpus"] = True
+            elif not synced_gpus:
+                raise ValueError("synced_gpus must be True when using FSDP.")
```
sidnarayanan, line 203: I think this is set here: https://github.com/huggingface/transformers/blob/79254c9b61ac2e1a6e0f698c38b1654159095e2e/src/transformers/generation/utils.py#L2014-L2015

src/ldp/alg/rollout.py
```diff
@@ -260,6 +343,9 @@ async def store_step(step: Transition):
         except CaughtError as e:
             # NOTE: This trajectory should not be used for regular training.
             # We save the last transition here for debugging, etc.
+            if log_exceptions_immediately:
+                logger.exception(f"Exception in rollout {traj_id}: {e.original_exc}")
+
```
sidnarayanan, line 348: Can you bring back `format_error_details`? It's needed for remote agents/envs that communicate over HTTP (i.e. the exception is the response body)

src/ldp/alg/rollout.py
```diff
@@ -153,61 +165,132 @@ async def rollout_with_args(idx: int, **rollout_kwargs):
...
+                    pbar.update(1)
+
+                    steps_in_traj = len(traj.steps)
+                    accumulated_steps[idx] += steps_in_traj
+
+                    # Check for exceptions in this trajectory
+                    if traj.steps and traj.steps[-1].metadata.get("exception"):
+                        exc_str: str = str(traj.steps[-1].metadata["exception"])[
+                            :500
+                        ].replace('"', "'")
```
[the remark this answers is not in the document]
kwanUm: We do, though I preferred a simple form of naive grouping. This worked well for me, as I believe those setting log_exceptions_immediately=False generally aim to reduce clutter. I can adjust it to display the full exception and remove grouping if needed.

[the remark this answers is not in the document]
sidnarayanan: In that case it might be better to rename`log_exceptions_immediately=False` to `summarize_exceptions=True`

src/ldp/graph/async_torch.py
```diff
@@ -19,6 +20,9 @@
         "Please run `pip install ldp[nn]`."
     ) from None
 
+
+logger = logging.getLogger(__name__)
```
sidnarayanan, line 24: Maybe I'm missing it in the diff, but is this logger used?

src/ldp/nn/agent/simple_local_agent.py
```diff
@@ -42,6 +46,10 @@ class AgentLMConfig(_LMConfig):
         ),
         validate_default=True,
     )
+    max_messages_token_count: int | None = Field(
+        default=None,
+        description="If set, raise an error if the total tokens in the trajectory exceed this value.",
```
sidnarayanan, line 51:
```suggestion
        description="If set, raise an error if the total tokens in the message history and tool description exceed this value.",
```

Since it includes tools, maybe we should call it `max_token_count` - I just don't want to call it a trajectory, since in principle it is just (state, observation) [e.g. for SeqQA we don't pass the full trajectory]

src/ldp/nn/agent/simple_local_agent.py
```diff
@@ -112,8 +122,30 @@ async def get_asv(
 
         # Update state messages with result and return the new state
         next_state.messages = [*next_state.messages, result.value]
+        self._validate_token_count(next_state.messages, next_state.tools)
```
sidnarayanan, line 125: Do we want to validate the output or the input? Why throw an error if the output is over the limit?

src/ldp/nn/agent/simple_local_agent.py
```diff
@@ -112,8 +122,30 @@ async def get_asv(
...
+            model=self.llm_model.model,
+            messages=messages_for_tokenizer,
+            tools=tools_for_tokenizer,  # type: ignore[arg-type]
+        )
+        if total_tokens > self.llm_model.max_messages_token_count:
+            logger.error(
+                f"Token limit exceeded for trajectory: {total_tokens} > {self.llm_model.max_messages_token_count}"
+            )
+            raise ValueError(
+                f"Token limit exceeded for trajectory: {total_tokens} > {self.llm_model.max_messages_token_count}"
```
sidnarayanan, line 146: Should remove trajectory reference here

sidnarayanan (approved):

PR #235, opened 2025-02-19
Fix aviary typing

maykcaldas:
https://github.com/Future-House/aviary/pull/188 changed FunctionInfo.parameters typing. This PR fixed downstream typing errors.


ldp/nn/graph/llm_call_op.py
```diff
@@ -97,16 +97,20 @@ def prep_tools_for_tokenizer(tools: list[Tool] | None) -> list[dict] | None:
                 "name": tool.info.name,
                 "description": tool.info.description,
                 "parameters": {
-                    "type": tool.info.parameters.type,
+                    "type": tool.info.parameters.type
+                    if tool.info.parameters
+                    else "object",
```
jamesbraza, line 102: Let's just blow up here if `tool.info.parameters` is `None` with a `NotImplementedError`

We don't need to handle Gemini for this for now

maykcaldas: Is gemini the only reason why we made this change? I added a generic message error saying that `Cannot serialize tools with no parameters.`

jamesbraza: Gemini is currently the only reason, but there can be other models. Good thinking making the message more generic

ldp/nn/graph/llm_call_op.py
```diff
@@ -92,8 +92,12 @@ def prep_tools_for_tokenizer(tools: list[Tool] | None) -> list[dict] | None:
             return None
 
         # TODO: should be able to switch to tool.info.model_dump() here
-        return [
-            {
+        tools_list = []
+        for tool in tools:
+            if tool.info.parameters is None:
+                raise NotImplementedError("Cannot serialize tools with no parameters.")
```
jamesbraza, line 98:
```suggestion
                raise NotImplementedError("Didn't yet handle serializing tools without parameters.")
```

It's not that we can't, it's that we didn't add this yet

jamesbraza (approved):

PR #238, opened 2025-02-20
Implements LMI

maykcaldas:


.github/workflows/tests.yml
```diff
@@ -37,19 +37,43 @@ jobs:
           enable-cache: true
       - run: uv python pin ${{ matrix.python-version }}
       - run: uv sync --python-preference=only-managed
-      - run: uv run refurb ldp tests
-      - run: uv run pylint ldp tests
+      - uses: pre-commit/action@v3.0.1
```
jamesbraza, line 40: Let's edit the `pre-commit` workflow, this is the `lint` workflow

.pre-commit-config.yaml
```diff
@@ -50,7 +54,7 @@ repos:
     rev: v2.3.0
     hooks:
       - id: codespell
-        additional_dependencies: [".[toml]"]
+        additional_dependencies: [".[toml,yaml]"]
```
jamesbraza, line 57: Why are we giving `codespell` YAML? The config is housed in the `pyproject.toml`

packages/lmi/README.md
```diff
@@ -0,0 +1,275 @@
+[![Python 3.11+](https://img.shields.io/badge/python-3.11+-blue.svg?style=plastic)]()
+[![tests](https://github.com/Future-House/llm-client/actions/workflows/test.yaml/badge.svg?style=plastic)](https://github.com/Future-House/ldp/tree/main/packages/llmi)
+[![PyPI version](https://badge.fury.io/py/lmi.svg?style=plastic)](https://badge.fury.io/py/lmi)
+
```
jamesbraza, line 4: Should we add the "TOC" directive here?

```md
<!--TOC-->

<!--TOC-->
```

packages/lmi/README.md
```diff
@@ -0,0 +1,275 @@
...
+```python
+from lmi import LiteLLMModel
+from aviary import Message
+
+llm = LiteLLMModel()
+
+messages = [
+    Message(content="What is the meaning of life?"),
+]
+
```
jamesbraza, line 28:
```suggestion
messages = [Message(content="What is the meaning of life?")]
```

More readable

pyproject.toml
```diff
@@ -210,6 +214,7 @@ load-plugins = [
 # Disable the message, report, category or checker with the given id(s).
 disable = [
     "arguments-differ",  # Ops intentionally differ arguments
+    "assignment-from-no-return",  # Don't care to enforce this
```
jamesbraza, line 217: What makes you do this? This seems like a good rule

pyproject.toml
```diff
@@ -220,6 +225,7 @@ disable = [
     "function-redefined",  # Rely on mypy no-redef for this
     "import-outside-toplevel",  # Rely on ruff PLC0415 for this
     "invalid-name",  # Don't care to enforce this
+    "keyword-arg-before-vararg",  # Don't care to enforce this
```
jamesbraza, line 228:
```suggestion
    "keyword-arg-before-vararg",  # Rely on ruff B026 for this
```

pyproject.toml
```diff
@@ -287,7 +296,7 @@ filterwarnings = [
 # files or test ids are given in the command line when executing pytest from the rootdir
 # directory. File system paths may use shell-style wildcards, including the recursive **
 # pattern.
-testpaths = ["src", "tests"]
+testpaths = ["packages/lmi/tests", "src", "tests"]
```
jamesbraza, line 299:
```suggestion
testpaths = ["packages/lmi", "src", "tests"]
```

We want to do doctests in `lmi` too

.pre-commit-config.yaml
maykcaldas: It was finding typos in the cassettes so I just removed it. Should I skip the cassettes here?
```
[tool.codespell]
check-filenames = true
check-hidden = true
ignore-words-list = "astroid,ser"
skip = "**tests/cassettes/*"
```

(I mean adding a skip for `packages/lmi/tests/cassete`)

packages/lmi/README.md
maykcaldas: We could also pass it directly to `call_single` I wanted to keep it separated to improve readability in the doc

pyproject.toml
maykcaldas: It is raising an error here:

https://github.com/Future-House/ldp/blame/main/ldp/nn/handlers/transformer_handler.py#L870-L874

I removed because it wasn't a problem in ldp before, so I assumed it was being ignored somehow

.pre-commit-config.yaml
jamesbraza: So this is telling `pre-commit` to `pip install codespell[toml,yaml]` during set up.

Yeah correct just make changes to the `codespell` config in `pyproject.toml`

pyproject.toml
jamesbraza: Just add `# pylint: disable= assignment-from-no-return ` to that specific line, don't disable globally

I think there may be a real bug with that `transformer_handler.py` line

packages/lmi/README.md
maykcaldas: Good idea! Done!

jamesbraza (approved):
Nice work here!

PR #240, opened 2025-02-21
Changed install name of lmi to fhlmi

maykcaldas:


ludomitch (approved):

.github/workflows/publish.yml
```diff
@@ -19,15 +19,15 @@ jobs:
...
       - name: Download built artifact to dist/
         uses: actions/download-artifact@v4
         with:
-          name: ${{ steps.build-lmi.outputs.artifact-name }}
+          name: ${{ steps.build-fhlmi.outputs.artifact-name }}
           path: dist
       - uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          password: ${{ secrets.PYPI_API_TOKEN }}
+          password: ${{ secrets.PYPI_LMI_API_TOKEN }}
```
jamesbraza, line 33: Why do we name this LMI token? Notably this will also publish LDP, so I don't think we should rename this

packages/lmi/pyproject.toml
```diff
@@ -32,20 +32,21 @@ dependencies = [
...
 ]
 description = "A client to provide LLM responses for FutureHouse applications."
-dynamic = ["version"]
 license = {file = "LICENSE"}
-name = "lmi"
+name = "fhlmi"
 readme = "README.md"
 requires-python = ">=3.11"
+# dynamic = ["version"]
+version = "0.0.1"
```
jamesbraza, line 40: Was this just for prototyping?

These packages are coversioned, we should keep them the same

PR #244, opened 2025-02-23
Fixed bad path in `publish.yml`

jamesbraza:
https://github.com/Future-House/ldp/pull/242 should have used `packages/lmi`


ludomitch (approved):

maykcaldas (approved):
mb, was the automatic substitution. Thanks for that fix!

PR #249, opened 2025-02-26
Defaults llm calls to use `include_reasoning` if using a deepseek model

maykcaldas:
This PR defaults `lmi` to using the `include_reasoning` attribute. Currently, it checks explicitly for deep-seek models because this is the only one supported by `Litellm` that produces reasoning so far.

This will be generalized once a better check is developed upstream: https://github.com/BerriAI/litellm/issues/8765


packages/lmi/src/lmi/llms.py
```diff
@@ -246,6 +246,19 @@ async def call(  # noqa: C901, PLR0915
         if n < 1:
             raise ValueError("Number of completions (n) must be >= 1.")
 
+        # NOTE: Checking if the model is a deepseek-r1-like model so that we can pass include_reasoning=True
+        # This hard-coded check will be removed once we have a better way to check if the model supports reasoning
+        # See: https://github.com/BerriAI/litellm/issues/8765
+        # Only reasoning with deepseek-r1 model is supported so far
+        if "deepseek" in self.name:
```
jamesbraza, line 253: Let's use `litellm.get_llm_provider` paired with `litellm.LlmProviders` instead of string literals

packages/lmi/src/lmi/llms.py
```diff
@@ -634,12 +647,6 @@ async def acompletion(self, messages: list[Message], **kwargs) -> list[LLMResult
     async def acompletion_iter(
         self, messages: list[Message], **kwargs
     ) -> AsyncIterable[LLMResult]:
-        if kwargs.get("include_reasoning"):
```
jamesbraza, line 637: Why remove this? `litellm` didn't fix the parent issue?

maykcaldas: I moved it to `call` so that all the checkings for `deep seek` and whether to use `include_reasoning` are in the same place

jamesbraza: Ah I see, resolving this one

packages/lmi/tests/test_llms.py
```diff
@@ -668,6 +668,58 @@ async def test_empty_tools(self, tools: list | None, model_name: str) -> None:
...
+                content="Think deeply about the following question and answer it.",
+            ),
+            Message(content="What is the meaning of life?"),
+        ]
+        results = await llm.call(messages)
+        for result in results:
+            assert result.reasoning_content
+
+        outputs: list[str] = []
+        results = await llm.call(messages, callbacks=[outputs.append])
```
jamesbraza, line 701: Do we need the `callbacks` here? It seems like we don't

packages/lmi/tests/test_llms.py
```diff
@@ -668,6 +668,58 @@ async def test_empty_tools(self, tools: list | None, model_name: str) -> None:
...
+        messages = [
+            Message(
+                role="system",
+                content="Think deeply about the following question and answer it.",
+            ),
+            Message(content="What is the meaning of life?"),
+        ]
+        results = await llm.call(messages)
+        for result in results:
+            assert result.reasoning_content
```
jamesbraza, line 698:
```
        results = await llm.call(messages)
        assert results
        for result in results:
            assert result.reasoning_content
```

Do we run the risk of empty `results` here? If we do, we should protect against it

packages/lmi/tests/test_llms.py
maykcaldas: It's because in `call` we have a path that uses `acompletion` -- if there's no `callback`. And we use `acompletion_iter` if we have. I want to test both paths.

packages/lmi/tests/test_llms.py
maykcaldas: I don't think we do. Worst case, `call` returns a `list` of `LLMResults` with empty strings. I'll add the protection for guarantee future implementations

packages/lmi/tests/test_llms.py
jamesbraza: Okay gotchu. Yeah them maybe let's assert on the contents of `outputs` so that way the use of the callback is not dead code

packages/lmi/tests/test_llms.py
```diff
@@ -668,6 +668,60 @@ async def test_empty_tools(self, tools: list | None, model_name: str) -> None:
...
+        assert isinstance(results, list)
+        assert isinstance(results[0], LLMResult)
+        assert results[0].reasoning_content
+
+        outputs: list[str] = []
+        with pytest.raises(
+            NotImplementedError,
+            match=r"Reasoning with OpenRouter is not supported in streaming mode.*",
+        ):
+            await llm.call(messages, callbacks=[outputs.append])
```
jamesbraza, line 722: Can we remove `callbacks` from here? Or can we use it?

It feels weird to have callbacks that are unused. If trying to test some control flow with this, then maybe add a comment stating that

PR #253, opened 2025-02-28
Removed NotImplementedError when using OpenRouter/deepseek with `stream=True`

maykcaldas:
The motivation for this change is to avoid breaking usage downstream. 
Even if using openrouter/deepseek with callbacks, the model should be able to return a result.


packages/lmi/tests/test_llms.py
```diff
@@ -716,15 +716,15 @@ async def test_openrouter_reasoning(self) -> None:
...
-        ):
-            # NOTE: We invoke the call with a callback to test the streaming
-            #  from LiteLLMModel.acompletion_iter().
-            # Because it is still not supported for OpenRouter models, we test
-            #  that it raises a NotImplementedError.
-            await llm.call(messages, callbacks=[outputs.append])
+        # NOTE: We invoke the call with a callback to test the streaming
+        #  from LiteLLMModel.acompletion_iter().
+        # Because it is still not supported for OpenRouter models, we assert
+        #  that the reasoning content is not present in the result.
```
jamesbraza, line 722: Can you somehow check for reasoning here, with `pytest.xfail`

packages/lmi/src/lmi/llms.py
```diff
@@ -256,12 +256,13 @@ async def call(  # noqa: C901, PLR0915
         # Only reasoning with deepseek-r1 model is supported so far
         if LlmProviders.DEEPSEEK.value in get_llm_provider(self.name):
             chat_kwargs["include_reasoning"] = True
-        if LlmProviders.OPENROUTER.value in get_llm_provider(self.name) and callbacks:
-            raise NotImplementedError(
-                "Reasoning with OpenRouter is not supported in streaming mode."
-                "https://github.com/BerriAI/litellm/issues/8631"
-                "Consider using LiteLLMModel(name='deepseek/deepseek-reasoner') instead."
-            )
+        # if LlmProviders.OPENROUTER.value in get_llm_provider(self.name) and callbacks:
```
jamesbraza, line 259: I would rather just remove this than keep it, it will be in `git` history

packages/lmi/tests/test_llms.py
maykcaldas: I added `test_openrouter_stream_reasoning_fails`. Is that enough?

packages/lmi/tests/test_llms.py
```diff
@@ -716,15 +716,32 @@ async def test_openrouter_reasoning(self) -> None:
...
+        results = await llm.call(messages, callbacks=[outputs.append])
+        assert isinstance(results, list)
+        assert isinstance(results[0], LLMResult)
+        assert outputs
+
+    @pytest.mark.vcr(match_on=[*VCR_DEFAULT_MATCH_ON, "body"])
+    @pytest.mark.asyncio
+    @pytest.mark.xfail(
+        reason="LiteLLM still doesn't support OpenRouter reasoning with streaming"
+    )
```
jamesbraza, line 734: Can you put an exception type to ensure this is accurately aimed? Think how we provide an exception type to `pytest.raises`

packages/lmi/tests/test_llms.py
jamesbraza: Sure will comment there, resolving this one

packages/lmi/tests/test_llms.py
```diff
@@ -716,15 +716,32 @@ async def test_openrouter_reasoning(self) -> None:
...
+    async def test_openrouter_stream_reasoning_fails(self) -> None:
+        llm = LiteLLMModel(name="openrouter/deepseek/deepseek-r1", config={"n": 1})
+        messages = [
+            Message(content="What is the meaning of life?"),
+        ]
+        outputs: list[str] = []
+        results = await llm.call(messages, callbacks=[outputs.append])
```
jamesbraza, line 741: Do we need `callbacks` here, when we don't use `outputs`? I feel like we should not be propagating this everywhere

maykcaldas: I merged the tests for deepseek and openrouter. Then we don't need to spread the dummy callback everywhere. What do you think?

jamesbraza (approved):
Nice work! 🥳

PR #258, opened 2025-03-04
Enable set RPM rate limiter

lingyaochu:
- Add a `request_limited` decorator to set RPM(request per minute) limit.
- Add a simple test for testing RPM feature using deepseek api.


packages/lmi/tests/test_rpm.py
```diff
@@ -0,0 +1,131 @@
...
+
+async def test_rpm_limit():
+    """Test the request rate limit (RPM) functionality"""
+    messages = [Message(role="user", content="Hi")]
+    results = {}
+    
+    # Test request time without rate limit
+    print("\n=== Testing without rate limit ===")
+    model_no_limit = LiteLLMModel(
+        name="deepseek/deepseek-chat",
```
jamesbraza, line 19: Do you mind doing your tests with `lmi.CommonLLMNames.OPENAI_TEST` to keep costs down? Or, if you're using DeepSeek for a particular reason here, can you add a comment to the test explaining why

packages/lmi/tests/test_rpm.py
```diff
@@ -0,0 +1,131 @@
...
+    
+    status = await GLOBAL_LIMITER.rate_limit_status()
+    print("\nCurrent limit status:", status)
+    
+    return results
+
+if __name__ == "__main__":
+    from dotenv import load_dotenv
+    
+    load_dotenv()
```
jamesbraza, line 129: Why do you have `load_dotenv` for the test? Is it to load `DEEPSEEK_API_KEY`?

If so, when you run `pytest` you'll need `DEEPSEEK_API_KEY=abc123 pytest -n auto`

packages/lmi/tests/test_rpm.py
```diff
@@ -0,0 +1,131 @@
...
+                    "max_tokens": 4096,
+                }
+            }]
+        }
+    )
+    
+    results["No Limit"] = {}
+    for req_count in [2, 3, 4]:
+        print(f"\nTesting time to complete {req_count} requests without limit...")
+        start_time = time.time()
```
jamesbraza, line 35: Let's use `time.perf_counter` to be more precise, on Windows `time.sleep` can be pretty coarse

jamesbraza (commented):
Interesting PR! I threw up some comments and can you get `pre-commit` and `lint` to pass

packages/lmi/tests/test_rpm.py
maykcaldas: Following up on this, when you run pytest, we have a `fixture` in `conftest.py` to load the `.env:
https://github.com/Future-House/ldp/blob/main/packages/lmi/tests/conftest.py#L24

You should be able to run the tests just by having a `.env` with the secrets on the root of the project

packages/lmi/tests/test_rpm.py
lingyaochu: Hi @jamesbraza,

Thanks for the suggestion! I understand the concern about costs.

The reason I used `deepseek/deepseek-chat` in `test_rpm.py` initially was for local testing purposes. Accessing OpenAI APIs can be challenging from China, and I didn't have immediate access during the initial development and testing phase.

Now that I've verified the code is working as expected with the `deepseek` model, I'm happy to switch it back to use `lmi.CommonLLMNames.OPENAI_TEST` as you suggested. I will made the change to use the OpenAI model, to be ready for future test.

Thanks again for your attention to detail!

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -47,9 +47,23 @@
 # the dynamic IP of the machine will be used to limit the rate of requests, otherwise the
 # user input machine_id will be used.
 
+# Default rate limits for various LLM providers
+OPENAI_DEFAULT_RPM = RateLimitItemPerMinute(
+    3500, 1
+)  # OpenAI default RPM for most models
```
jamesbraza, line 53:
```suggestion
OPENAI_DEFAULT_RPM = RateLimitItemPerMinute(  # OpenAI default RPM for most models
    3500, 1
)
```

packages/lmi/tests/test_rpm.py
```diff
@@ -0,0 +1,148 @@
...
+import asyncio
+import time
+
+import pytest
+
+from lmi.llms import CommonLLMNames, LiteLLMModel
+from lmi.types import Message
+
+
+def _setup_model_config(rpm_limit: int | None) -> tuple[dict, str]:
```
jamesbraza, line 16: Can we default `rpm_limit: int | None = None`

packages/lmi/tests/test_rpm.py
```diff
@@ -0,0 +1,148 @@
...
+                    "max_tokens": 4096,
+                },
+            }
+        ]
+    }
+
+    if rpm_limit is not None:
+        model_config["request_limit"] = {
+            "deepseek/deepseek-chat": f"{rpm_limit}/minute"
+        }  # type: ignore[assignment]
```
jamesbraza, line 41:
```suggestion
        model_config["request_limit"] = {  # type: ignore[assignment]
            "deepseek/deepseek-chat": f"{rpm_limit}/minute"
        }
```

packages/lmi/tests/test_rpm.py
```diff
@@ -0,0 +1,148 @@
...
+
+    return model_config, f"RPM={rpm_limit}" if rpm_limit else "No Limit"
+
+
+def _validate_rpm_results(results: dict[str, float]) -> None:
```
jamesbraza, line 46: So in unit testing, it's best to avoid helper functions, if they're only used once.

The idea is unit tests should have "locality of behavior" which means they can be read in one place, without reading helper functions.

It's okay if a unit test is a bit longer, if you're hitting line length issues from linters just disable them

packages/lmi/tests/test_rpm.py
```diff
@@ -0,0 +1,148 @@
...
+    try:
+        if is_concurrent:
+            concurrent_tasks = [model.call_single(messages) for _ in range(req_count)]
+            await asyncio.gather(*concurrent_tasks)
+        else:
+            for _ in range(req_count):
+                await model.call_single(messages)
+        results[test_name] = time.perf_counter() - start_time
+    except (ValueError, RuntimeError) as e:
+        pytest.fail(f"Rate-limited requests failed: {e!s}")
```
jamesbraza, line 122: So the test will fail if an `Exception` is raised. Imo there's no need to catch the exception just to manually fail the test.

Do you mind removing the try-except? It's easier to debug failing tests with a full stack trace from a normal exception

packages/lmi/tests/test_rpm.py
```diff
@@ -0,0 +1,148 @@
+"""Tests for request rate limit (RPM) functionality in LiteLLMModel.
```
jamesbraza, line 1: Can you move all of this to `test_rate_limiter.py`? Since the "software under test" is in `rate_limiter.py`

It's okay if `test_rate_limiter.py` has a huge line count

packages/lmi/src/lmi/llms.py
```diff
@@ -574,6 +617,7 @@ async def check_rate_limit(self, token_count: float, **kwargs) -> None:
                 **kwargs,
             )
 
+    @request_limited
```
jamesbraza, line 620: How do you feel about combining `request_limited` with `rate_limited`?

Do you think it's possible and can make sense?

It's okay to break our signatures, we can just call it `@limited` then

jamesbraza (commented):
This is a nice PR, keep it up!

packages/lmi/tests/test_rpm.py
lingyaochu: Okay, I will move all the tests to `test_rate_limiter.py`.  While the test logic itself is quite straightforward, the task will involve more than just a simple move. I need to ensure these tests are properly integrated and aligned with the conventions already established in `test_rate_limiter.py`.  This might require some time to do correctly.

packages/lmi/src/lmi/llms.py
lingyaochu: Thank you for the suggestion. However, I’m hesitant to combine `request_limited` with `rate_limited` into a single decorator for two key reasons:

- User Disruption: Changing established conventions would force users to adapt to a new pattern, which may cause unnecessary friction.

- Implementation Burden: Refactoring all existing usages would require significant time and effort—resources I currently can’t allocate due to higher-priority tasks.

Unless this change is critical for immediate goals, I recommend keeping the current structure to avoid compounding technical debt and workload. Let me know if further discussion is needed.

jamesbraza: Alright sounds good. Fwiw in this repo we are in a "[zero ver](https://0ver.org/)" place where we don't mind doing breaks every so often.

That being said, yeah refactoring existing usages can be burdensome

Thanks for looking into it, resolving this one

packages/lmi/tests/test_rate_limiter.py
```diff
@@ -293,3 +325,64 @@ async def test_embedding_rate_limits(
...
+        is_concurrent: Whether to run requests concurrently.
+        llm_config_w_request_limits: llm config with request configuration parameters.
+    """
+    rate_limit_item = next(
+        iter(llm_config_w_request_limits["config"]["request_limit"].values())
+    )
+    req_request = rate_limit_item.amount
+    req_count = req_request + 1
+
+    messages = [Message.create_message(role="user", text=RATE_LIMITER_PROMPT)]
```
jamesbraza, line 345:
```suggestion
    messages = [Message(content=RATE_LIMITER_PROMPT)]
```

Easier just to directly instantiate, and default `role` is already `"user"`

jamesbraza (commented):
Can you get `pre-commit` to pass?  You can run it locally fwiw, no need for our CI to run

```bash
uv venv
source .venv/bin/activate
uv sync
pre-commit install
pre-commit run -a
```

lingyaochu:
``> Can you get `pre-commit` to pass? You can run it locally fwiw, no need for our CI to run
> 
> ```shell
> uv venv
> source .venv/bin/activate
> uv sync
> pre-commit install
> pre-commit run -a
> ```

I followed your instructions and found that the mypy check consistently throws errors. However, the flagged code is unrelated to my changes. For now, I've added type-ignore comments to bypass checking the affected sections, avoiding direct modifications to the original code.

packages/lmi/src/lmi/llms.py
```diff
@@ -385,7 +386,7 @@ async def call_single(
 P = ParamSpec("P")
 
 
-@overload
+@overload  # type: ignore[no-overload-impl]
```
jamesbraza, line 389: So this `mypy` error is here for a real error:

```python
@overload  # type: ignore[no-overload-impl]
def rate_limited(
    func: Callable[P, Coroutine[Any, Any, list[LLMResult]]],
) -> Callable[P, Coroutine[Any, Any, list[LLMResult]]]: ...


@overload
def rate_limited(
    func: Callable[P, AsyncIterable[LLMResult]],
) -> Callable[P, Coroutine[Any, Any, AsyncIterable[LLMResult]]]: ...


def request_limited(func):
```

This code is applying `typing.overload` for a `rate_limited` function onto `request_limited`, hence the `mypy` error. The `overload` should be directly over `rate_limited`.

The fix is to move `request_limited` below or above `rate_limited` entirely, and not split the `overload` away from the core function.

lingyaochu: @jamesbraza Thank you for the detailed review. I realize now that blindly adding `type: ignore` to bypass mypy errors—instead of addressing the root cause—was a serious oversight. Your explanation and pointing out the misplaced `@overload` were very helpful. I've removed the unnecessary type ignores and corrected the decorator placement as you suggested. My local pre-commit checks are now passing cleanly. I appreciate the guidance and will be more careful in the future!

packages/lmi/src/lmi/llms.py
```diff
@@ -441,6 +442,36 @@ async def rate_limited_generator() -> AsyncIterable[LLMResult]:
     return wrapper
 
 
+def request_limited(func):
```
jamesbraza, line 445: Can you look at `rate_limited`'s typing and propagate it here, both to `request_limited` and its inner `wrapper`?

Before this PR, the typing was inspectable with `mypy`. So with this code

```python
from typing import reveal_type

from lmi.llms import LiteLLMModel

model = LiteLLMModel()
reveal_type(model.acompletion)
```

Running `mypy` on this prints:

```none
note: Revealed type is "def (messages: builtins.list[aviary.message.Message], **kwargs: Any) -> typing.Coroutine[Any, Any, builtins.list[lmi.types.LLMResult]]"
```

With this PR, running `mypy` now prints:

```none
note: Revealed type is "Any"
```

So we see this change regresses the typing inspect-ability. Let's add typing here, then it's LGTM

packages/lmi/src/lmi/llms.py
jamesbraza: No problem, thanks for resolving it! We live and learn

packages/lmi/src/lmi/llms.py
lingyaochu: @jamesbraza 

Thanks for the feedback! I've added `@overload` to `request_limited` to mirror `rate_limited`'s type annotations, ensuring `mypy` now infers the correct types (verified in testing).

About combining the decorators: I had earlier decided against merging `rate_limited` and `request_limited`, but I'm now trying out the idea to see if it improves the code. If it works, I might include it in a separate PR later. For now, this PR only focuses on reliazation of request limit.

Let me know if anything needs adjusting!

jamesbraza: Nice work per typing, looks good!

And sounds good per merging them. Yeah there's really not many usages of `rate_limited` so far, so it should not be much work to merge them if you have time.

Thanks again for working through everything

jamesbraza (approved):
Nice work @lingyaochu

maykcaldas (commented):
That's a nice feature. Thanks for that @lingyaochu. 

I have only a minor question about the default for RATE_CONFIG. But the PR looks nice!

Can you also update the [Rate Limiting](https://github.com/Future-House/ldp/tree/main/packages/lmi#rate-limiting) section of the readme to show this new feature, please?

maykcaldas:
Can you also merge the `main` branch to see if it solves some CI errors?

packages/lmi/src/lmi/llms.py
```diff
@@ -565,6 +610,16 @@ def router(self) -> litellm.Router:
                 )
         return self._router
 
+    async def check_request_limit(self, **kwargs) -> None:
+        """Check if the request is within the request rate limit."""
+        if "request_limit" in self.config:
+            await GLOBAL_LIMITER.try_acquire(
+                ("request", self.name),
```
mskarlin, line 617: Can we change the namespace for this to be `client|request` -- as request I think is a bit too general, and this refers to request from a particular client.

maykcaldas (approved):
Check comment about the namespace for the default RATE_CONFIG that @mskarlin also pointed out. Besides that, it's good to me!

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -47,9 +47,23 @@
...
+
 RATE_CONFIG: dict[tuple[str, str | MatchAllInputs], RateLimitItem] = {
     ("get", CROSSREF_BASE_URL): RateLimitItemPerSecond(30, 1),
     ("get", SEMANTIC_SCHOLAR_BASE_URL): RateLimitItemPerSecond(15, 1),
+    ("client", "gpt-3.5-turbo"): OPENAI_DEFAULT_RPM,
+    ("client", "gpt-4"): RateLimitItemPerMinute(500, 1),  # Lower RPM for GPT-4
+    ("client", "gpt-4o"): RateLimitItemPerMinute(500, 1),  # Lower RPM for GPT-4o
+    ("client", "claude-3"): ANTHROPIC_DEFAULT_RPM,
+    ("client", "claude-3.5-sonnet"): ANTHROPIC_DEFAULT_RPM,
+    ("client", "gemini-pro"): GOOGLE_DEFAULT_RPM,
```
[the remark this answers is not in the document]
lingyaochu: I've changed all instances to "client|request". Additionally, I only added RPM configurations for a few models. There might be a need for another way to configure all LLM models based on their API provider's limits.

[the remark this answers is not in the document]
lingyaochu: @maykcaldas I've added the `request_limited` examples to the `README`. Could you take a look and confirm whether it adequately addresses the point about distinguishing it from rate_limited and if any further clarification is required?

lingyaochu:
I think I've found a potential bug in the `Monitoring Rate Limits` section of the README (linked [here](https://github.com/Future-House/ldp/tree/main/packages/lmi#monitoring-rate-limits)).  The example currently uses `gpt-4o`, but due to regional limitations, I've been testing with `deepseek/deepseek-chat` to verify `rate_limit` and `request_limit` functionality.

Here's the test code I'm using:

```python
from lmi.rate_limiter import GLOBAL_LIMITER
from lmi import LiteLLMModel
from aviary.core import Message
import asyncio

config = {
    "rate_limit": {
        "deepseek/deepseek-chat": "100/minute",  # 100 tokens per minute
    },
    "request_limit": {
        "deepseek/deepseek-chat": "5/minute",  # 5 requests per minute
    }
}

async def main():
    llm = LiteLLMModel(name="deepseek/deepseek-chat", config=config)
    results = await llm.call([Message(content="Hello, world!")])  # Consume some tokens
    status = await GLOBAL_LIMITER.rate_limit_status()
    print(status)

if __name__ == "__main__":
    asyncio.run(main())
```
However, the output from `GLOBAL_LIMITER.rate_limit_status()` is unexpected:
```python
{
    ('client|request', 'deepseek'): {
        'period_start': 1741796675.0483532,
        'n_items_in_period': 0,
        'period_seconds': 1,
        'period_name': 'second', 'period_cap': 3
        }, 
    ('client', 'deepseek'): {
        'period_start': 1741796675.0483532, 
        'n_items_in_period': 0, 
        'period_seconds': 60, 
        'period_name': 
        'minute', 'period_cap': 30000
        }
    }
```
Observations:
- The model name is incorrect. It returns 'deepseek' instead of the expected 'deepseek/deepseek-chat'
- The `rate_limit` and `request_limit` configurations don't seem to be applied. They're using the default values instead of the `100/minute` and `5/minute` I specified.
- 
I suspect this might be related to the `parse_key` method in `rate_limiter.py`. Currently, it uses `namespace, primary_key = key.split("/")[1:3]` to extract the namespace and key. This works for OpenAI, but for LiteLLM (and perhaps other providers) where the model name is formatted as `[provider]/[model]`, this splitting logic appears to be causing the incorrect parsing.

packages/lmi/README.md
```diff
@@ -124,8 +124,8 @@ Cost tracking is supported in two different ways:
 
 ### Rate limiting
 
-Rate limiting helps control the rate of tokens used to various services and LLMs. The rate limiter supports both in-memory and Redis-based storage for cross-process rate limiting.
-Currently, `lmi` only take into account the tokens used (hence, it only limits Tokens Per Minute (TPM)).
+Rate limiting helps regulate the usage of resources to various services and LLMs. The rate limiter supports both in-memory and Redis-based storage for cross-process rate limiting.
+Currently, `lmi` take into account the tokens used (the Token per Minute(TPM)) and the request handled (the Request per Minute (RPM)).
```
jamesbraza, line 128:
```suggestion
Currently, `lmi` take into account the tokens used (Tokens per Minute (TPM)) and the requests handled (Requests per Minute (RPM)).
```

packages/lmi/README.md
```diff
@@ -214,7 +224,10 @@ from aviary.core import Message
 config = {
     "rate_limit": {
         "gpt-4": "100/minute",  # 100 tokens per minute
-    }
+    },
+    "request_limit": {
+        "gpt-4": "5/minute",  # 5 request per minute
```
jamesbraza, line 229:
```suggestion
        "gpt-4": "5/minute",  # 5 requests per minute
```

packages/lmi/README.md
```diff
@@ -139,20 +139,30 @@ from lmi import LiteLLMModel
...
+
 2. Through the global rate limiter configuration:
 
 ```python
 from lmi.rate_limiter import GLOBAL_LIMITER
 
-GLOBAL_LIMITER.rate_config[("client", "gpt-4")] = "100/minute"
+GLOBAL_LIMITER.rate_config[("client", "gpt-4")] = "100/minute"  # 100 tokens per minute
+GLOBAL_LIMITER.rate_config[("client|request", "gpt-4")] = (
+    "5/minute"  # 5 request per minute
```
jamesbraza, line 160:
```suggestion
    "5/minute"  # 5 requests per minute
```

packages/lmi/README.md
```diff
@@ -139,20 +139,30 @@ from lmi import LiteLLMModel
...
+)
 ```
 
+With `client` and `client|request`, you can control both token consumption and request volume. You can also configure only one of them if needed.
```
jamesbraza, line 164:
```suggestion
With `client` we rate limit only token consumption,
and with `client|request` we rate limit both token consumption and request volume.
You can also configure only one of them if needed.
```

packages/lmi/README.md
```diff
@@ -224,13 +237,20 @@ status = await GLOBAL_LIMITER.rate_limit_status()
 
 # Example output:
 {
+    ("client|request", "gpt-4"): {
+        "period_start": 1234567890,
+        "n_items_in_period": 1,
```
jamesbraza, line 242: Can you explain in a comment what this `n_items_in_period` means for `client|request`?

Does it mean one request, one token, or both?

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -47,9 +47,22 @@
...
+
 RATE_CONFIG: dict[tuple[str, str | MatchAllInputs], RateLimitItem] = {
     ("get", CROSSREF_BASE_URL): RateLimitItemPerSecond(30, 1),
     ("get", SEMANTIC_SCHOLAR_BASE_URL): RateLimitItemPerSecond(15, 1),
+    ("client|request", "gpt-3.5-turbo"): RateLimitItemPerMinute(3500, 1),
+    ("client|request", "gpt-4o"): OPENAI_DEFAULT_RPM,
+    ("client|request", "gpt-4"): OPENAI_DEFAULT_RPM,
+    ("client|request", "claude-3"): ANTHROPIC_DEFAULT_RPM,
+    ("client|request", "claude-3.5-sonnet"): ANTHROPIC_DEFAULT_RPM,
+    ("client|request", "gemini-2.0-flash"): GOOGLE_DEFAULT_RPM,
```
maykcaldas, line 65: I'm not sure how these "client|request" work. Does it mean we rate limit either "client" *or* "request" with the same rate? Is it cumulative? If it's just setting up the same rate for both, we will never hit RPM (TPM >>> RPM). Can you just make sure these rates are correctly set?

maykcaldas (approved):

packages/lmi/src/lmi/rate_limiter.py
lingyaochu: > Can we change the namespace for this to be `client|request` -- as request I think is a bit too general, and this refers to request from a particular client.

As suggested by @mskarlin, I have changed the namespace for `request_limit` from `"request"` to `"client|request"`. The `"client|request"` namespace is a single string and does not imply an "or" relationship in Python. It specifically refers to rate limiting *requests*. If this causes any confusion, I can adjust the naming to make it clearer.

packages/lmi/README.md
lingyaochu: > > Can we change the namespace for this to be `client|request` -- as request I think is a bit too general, and this refers to request from a particular client.
> 
> As suggested by @mskarlin, I have changed the namespace for `request_limit` from `"request"` to `"client|request"`. The `"client|request"` namespace is a single string and does not imply an "or" relationship in Python. It specifically refers to rate limiting _requests_. If this causes any confusion, I can adjust the naming to make it clearer.

I made an explanation here. Maybe it's better to change the namespace for `request_limit` and `rate_limit` to reveal their actual function. Do you have any suggestions?

packages/lmi/README.md
```diff
@@ -139,20 +139,34 @@ from lmi import LiteLLMModel
 config = {
     "rate_limit": {
         "gpt-4": "100/minute",  # 100 tokens per minute
-    }
+    },
+    "request_limit": {
+        "gpt-4": "5/minute",  # 5 request per minute
```
maykcaldas, line 144:
```suggestion
        "gpt-4": "5/minute",  # 5 requests per minute
```

packages/lmi/src/lmi/rate_limiter.py
maykcaldas: Gotcha! Thanks for the explanation!

maykcaldas (approved):
Hey @lingyaochu , nice job here!

I just pointed out a last typo in the README. I am happy with the implementation 🤘🏻

packages/lmi/src/lmi/rate_limiter.py
jamesbraza: I also was confused a bit and interpreted `client|request` as an "or". I think `client.request` or `client>request` might make more sense.

However, since this code has convention with `f"get|{MATCH_MACHINE_ID}"` below, I guess we can just go with it for now.

jamesbraza (approved):
Nice work again! Merging this now, will cut a release next week

PR #263, opened 2025-03-07
Low resolution rate limits

mskarlin:
In distributed scenarios -- redis may get millions of tokens per minute as weights to the rate limiter. This causes serious performance issues, so we need the ability for the rate limiter to support lower resolutions, like counting every 1k tokens for example. This PR gives us the ability to do that. I also added a test and swapped us to using `perf_counter`.

### DIFF
--- a/packages/lmi/src/lmi/rate_limiter.py
+++ b/packages/lmi/src/lmi/rate_limiter.py
@@ -329,6 +329,13 @@ class GlobalRateLimiter:
             }
         return limit_status
 
+    def maybe_lower_weight_with_rate_resolution(
+        self, weight: int, rate_limit: RateLimitItem
+    ) -> int:
+        if hasattr(rate_limit, "resolution"):
+            return max(1, weight // rate_limit.resolution)
+        return weight
+
     async def try_acquire(
         self,
         namespace_and_key: tuple[str, str],
@@ -378,9 +385,13 @@ class GlobalRateLimiter:
 
         rate_limit = rate_limit or rate_limit_
 
-        if rate_limit.amount < weight and raise_impossible_limits:
+        adjusted_weight = self.maybe_lower_weight_with_rate_resolution(
+            weight, rate_limit
+        )
+
+        if rate_limit.amount < adjusted_weight and raise_impossible_limits:
             raise ValueError(
-                f"Weight ({weight}) > RateLimit ({rate_limit}), cannot satisfy rate"
+                f"Weight ({adjusted_weight}) > RateLimit ({rate_limit}), cannot satisfy rate"
                 " limit."
             )
         while True:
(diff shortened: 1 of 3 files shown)


packages/lmi/tests/test_rate_limiter.py
```diff
@@ -293,3 +294,64 @@ async def test_embedding_rate_limits(
...
+    # try again with a new entry, which should also be under the limit
+    namespace_and_key_2 = ("client", "test-client2")
+
+    await limiter.try_acquire(
+        namespace_and_key_2,
+        rate_limit=rate_limit,
+        weight=4500,  # Should be adjusted to 4.5, which is under the limit of 5
+        acquire_timeout=5.0,
+    )
+    second_acquire_time = time.perf_counter() - start_time
```
maykcaldas, line 338: Isn't these two tests testing the same thing? The only difference is that the first one pass weight using a variable and it's hardcoded in the second one

maykcaldas (commented):
Just a small comment. But I think this is good

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -329,6 +329,13 @@ async def rate_limit_status(self):
             }
         return limit_status
 
+    def maybe_lower_weight_with_rate_resolution(
+        self, weight: int, rate_limit: RateLimitItem
+    ) -> int:
+        if hasattr(rate_limit, "resolution"):
+            return max(1, weight // rate_limit.resolution)
+        return weight
```
jamesbraza, line 337: Do you mind moving this to a method of `RateLimitItem`? We aren't using `self` here, so it doesn't need to be on `GlobalRateLimiter`

The below would become:

```python
        adjusted_weight = rate_limit.maybe_lower_weight_via_resolution(weight)
```

Which feels cleaner

packages/lmi/src/lmi/types.py
```diff
@@ -109,3 +112,18 @@ def provider(self) -> str:
     def get_supported_openai_params(self) -> list[str] | None:
         """Get the supported OpenAI parameters for the model."""
         return litellm.get_supported_openai_params(self.model)
+
+
+# Following classes allow you to adjust the "resolution" of weights
+# for counting tokens this increases performance, and lowers your
+# resolution to the nearest multiple of the resolution.
+class RateLimit1kItemPerMinute(RateLimitItemPerMinute):
+    resolution: int = 1_000
```
jamesbraza, line 121: Do you mind documenting the units as tokens?

packages/lmi/src/lmi/rate_limiter.py
jamesbraza: Oh if it's not a custom class of ours, don't subclass for this. Just do a `staticmethod` or free function here

Or just totally ignore this comment, lol

packages/lmi/src/lmi/types.py
jamesbraza: Ah I see, resolving this one

jamesbraza (approved):
Nice work here!

PR #264, opened 2025-03-11
Simplified `refurb` in CI after #238 fixed `mypy` config

jamesbraza:
I realized this today, and it's awesome


maykcaldas (approved):
Clean code goes brrr!

PR #265, opened 2025-03-11
Updated litellm to support openrouter reasoning on streaming

maykcaldas:
https://github.com/BerriAI/litellm/releases/tag/v1.63.5-nightly released the fix for reasoning on streaming with openrouter. This PR updates our test


uv.lock
```diff
@@ -3647,7 +3655,7 @@ wheels = [
 
 [[package]]
 name = "torch"
-version = "2.5.1"
+version = "2.6.0"
```
jamesbraza, line 3658: You'll wanna just do `uv lock` here not `uv lock --upgrade`, because this isn't `torch==2.6.0` compatible yet

jamesbraza (approved):

uv.lock
maykcaldas: Oh, this was `uv sync`. Thanks for the fix!

PR #266, opened 2025-03-12
Used PREFIX from RedisStorage instead of using it from the instance

maykcaldas:
`limits` changed how to get the `PREFIX` from `RedisStorage` on version 4.2.
We're getting it directly from the class now

### DIFF
--- a/packages/lmi/src/lmi/rate_limiter.py
+++ b/packages/lmi/src/lmi/rate_limiter.py
@@ -285,7 +285,7 @@ class GlobalRateLimiter:
             while cursor:
                 cursor, keys = await client.scan(
                     int(cursor),
-                    match=f"{storage.PREFIX}*",
+                    match=f"{storage.bridge.PREFIX}*",
                     count=cursor_scan_count,
                 )
                 matching_keys.extend(list(keys))
--- a/packages/lmi/src/lmi/rate_limiter.py
+++ b/packages/lmi/src/lmi/rate_limiter.py
@@ -284,7 +284,7 @@ class GlobalRateLimiter:
             while cursor:
                 cursor, keys = await client.scan(
                     int(cursor),
-                    match=f"{self.storage.PREFIX}*",
+                    match=f"{RedisStorage.PREFIX}*",
                     count=cursor_scan_count,
                 )
                 matching_keys.extend(list(keys))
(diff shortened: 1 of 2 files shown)


packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -284,7 +284,7 @@ async def get_rate_limit_keys(
             while cursor:
                 cursor, keys = await client.scan(
                     int(cursor),
-                    match=f"{self.storage.PREFIX}*",
+                    match=f"{RedisStorage.PREFIX}*",
```
jamesbraza, line 287: `self.storage` is `RedisStorage | MemoryStorage` and:

```
In [3]: MemoryStorage.PREFIX
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 MemoryStorage.PREFIX

AttributeError: type object 'MemoryStorage' has no attribute 'PREFIX'
```

Can you add a comment explaining why we `match` on just `RedisStorage`?

maykcaldas: We check on that a few lines earlier:
https://github.com/Future-House/ldp/blob/ae84e90114b02a0b88bcce2272524b2420a35fe7/packages/lmi/src/lmi/rate_limiter.py#L274-L290

jamesbraza: Ah then what we should do is let `mypy` know that:

```
        storage = self.storage
        if not isinstance(storage, RedisStorage):
            raise NotImplementedError(
                "get_rate_limit_keys only works with RedisStorage."
``

And have this line be `{storage.PREFIX}`

Comment (maykcaldas) [reply packages/lmi/src/lmi/rate_limiter.py]: You mean making it local makes mypy understand that it could not be changed elsewhere?
Updated it

Comment (jamesbraza) [reply packages/lmi/src/lmi/rate_limiter.py]: If you:

1. Use `typing.reveal_type`
2. Run `mypy` locally
3. You will see `mypy` can detect the type narrowing

File: packages/lmi/src/lmi/rate_limiter.py [hunk 2]
```diff
@@ -271,7 +271,8 @@ async def get_rate_limit_keys(
         if not (host and port):
             raise ValueError(f"Invalid Redis URL: {redis_url}.")
 
-        if not isinstance(self.storage, RedisStorage):
+        storage = self.storage
```
jamesbraza, line 274: Can you do `uv lock --upgrade-package limits`

jamesbraza (approved):

PR #267, opened 2025-03-19
WIP - handling annoying error in another coroutine

kwanUm:


src/ldp/graph/async_torch.py
```diff
@@ -106,6 +106,10 @@ async def __call__(self, **kwargs):
 
         while True:
             async with self._lock:
+                if self._exception_raised is not None:
+                    logger.info("Exception raised in another coroutine")
```
jamesbraza, line 110: Can we make this `warning` or `error` level?

PR #269, opened 2025-03-21
Moved a few `tree` imports to `TYPE_CHECKING`

jamesbraza:
I was investigating if we can make this an optional dependency, turns out we can't.

### DIFF
--- a/src/ldp/graph/loss_ops.py
+++ b/src/ldp/graph/loss_ops.py
@@ -2,14 +2,13 @@
 
 from typing import TYPE_CHECKING
 
-import tree
-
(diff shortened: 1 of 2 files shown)


maykcaldas (approved):
lgtm.

Can we just pin dm-tree to 0.1.8, where we know there is a wheel? Or do we need 0.1.9?

jamesbraza:
> Can we just pin dm-tree to 0.1.8, where we know there is a wheel

I also considered this. What should happen is the caller (in this case our platform) should do that pinning, I don't think LDP should pin `dm-tree` merely because there's not macOS wheels for it

PR #270, opened 2025-04-04
Updated lmi README to be explicit about multi-modality support

maykcaldas:


packages/lmi/README.md
```diff
@@ -69,7 +69,10 @@ Because these are the only methods that communicate with the chosen LLM provider
 
 #### LLMModel
 
-An `LLMModel` implements `call`, which receives a list of `aviary.Message`s and returns a list of `LLMResult`s. `LLMModel.call` can receive callbacks, tools, and output schemas to control its behavior, as better explained below.
+An `LLMModel` implements `call`, which receives a list of `aviary.Message`s and returns a list of `LLMResult`s.
```
jamesbraza, line 72:
```suggestion
An `LLMModel` implements `call`, which receives a list of `aviary` `Message`s and returns a list of `LLMResult`s. 
```

Note that `aviary.Message` is an `AttributeError`

jamesbraza (approved):

packages/lmi/README.md
maykcaldas: Yeah, thanks for catching that. We have already had this discussion before 😅
Fixed

PR #277, opened 2025-04-15
[WIP] Removed the limitation of having only one model_name on LLMModels in lmi

maykcaldas:


mskarlin (approved):
I know you're adding tests, but this LGTM and I can't find any use cases where this is needed so far.

packages/lmi/tests/test_llms.py
```diff
@@ -125,6 +125,66 @@ async def test_call(self, config: dict[str, Any]) -> None:
...
+                            },
+                        },
+                    ],
+                },
+                id="multiple-models",
+            ),
+        ],
+    )
+    @pytest.mark.vcr(match_on=[*VCR_DEFAULT_MATCH_ON, "body"])
+    async def test_call_w_multiple_models(self, config: dict[str, Any]) -> None:
```
jamesbraza, line 166: Might be cool to generalize this test a bit to account for an actual fallback taking place

PR #279, opened 2025-04-16
Allow agent fallback configuration

mskarlin:
We re-use our agent's input `llm_model` configuration in both the `LLMModel` instantiation and in our `call` kwargs. These have some overlapping, but otherwise different arguments. We could split these into different configs, but this would be a big LDP change because it changes the call signature on an LLM op's forward function. 

So, we're just making this such that an Agent can actually define an llm configuration and not break if any inputs are being used for the router config which aren't supported as `call` kwargs. 

As a little clean up -- 🧹  we had a couple places with mutated kwargs, I am using deep copies to fix this.

### DIFF
--- a/packages/lmi/src/lmi/llms.py
+++ b/packages/lmi/src/lmi/llms.py
@@ -11,6 +11,7 @@ __all__ = [
 
 import asyncio
 import contextlib
+import copy
 import functools
 import json
 import logging
@@ -231,7 +232,7 @@ class LLMModel(ABC, BaseModel):
         output_type: type[BaseModel] | TypeAdapter | JSONSchema | None = None,
         tools: list[Tool] | None = None,
         tool_choice: Tool | str | None = TOOL_CHOICE_REQUIRED,
-        **chat_kwargs,
+        **kwargs,
     ) -> list[LLMResult]:
         """Call the LLM model with the given messages and configuration.
 
@@ -247,6 +248,11 @@ class LLMModel(ABC, BaseModel):
         Raises:
             ValueError: If the LLM type is unknown.
         """
+        chat_kwargs = copy.deepcopy(kwargs)
+        # if using the config for an LLMModel,
+        # there may be a nested 'config' key
+        # that can't be used by chat
+        chat_kwargs.pop("config")
         n = chat_kwargs.get("n") or self.config.get("n", 1)
         if n < 1:
             raise ValueError("Number of completions (n) must be >= 1.")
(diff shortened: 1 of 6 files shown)


tests/test_rollouts.py
```diff
@@ -82,6 +83,70 @@ async def test_rollout(training: bool) -> None:
     assert all(v == 2 for v in callback.fn_invocations.values())
 
 
+# @pytest.mark.vcr
```
maykcaldas, line 86: I think we want to turn this vcr back on

maykcaldas (approved):

packages/lmi/src/lmi/llms.py
```diff
@@ -527,13 +533,20 @@ class LiteLLMModel(LLMModel):
 
     @model_validator(mode="before")
     @classmethod
-    def maybe_set_config_attribute(cls, data: dict[str, Any]) -> dict[str, Any]:
+    def maybe_set_config_attribute(cls, input_data: dict[str, Any]) -> dict[str, Any]:
```
jamesbraza, line 536: We have hit this issue before where our validators mutate the inputs, leaving side effects. We should type the `input_data` with `Mapping` to prevent this issue

tests/test_rollouts.py
```diff
@@ -82,6 +83,70 @@ async def test_rollout(training: bool) -> None:
...
+        "llm_model": {
+            "name": "openai/gpt-4o-mini",
+            "config": {
+                "model_list": AGENT_MODEL_LIST,
+                "router_kwargs": AGENT_ROUTER_KWARGS,
+            },
+        }
+    }
+    if fallback:
+        AGENT_CONFIG["llm_model"]["config"]["fallbacks"] = [  # type: ignore[index]
```
jamesbraza, line 124: We should assert the config is not mutated after the `SimpleAgent` is constructed

packages/lmi/src/lmi/llms.py
```diff
@@ -247,6 +248,11 @@ async def call(  # noqa: C901, PLR0915
         Raises:
             ValueError: If the LLM type is unknown.
         """
+        chat_kwargs = copy.deepcopy(kwargs)
+        # if using the config for an LLMModel,
+        # there may be a nested 'config' key
+        # that can't be used by chat
+        chat_kwargs.pop("config", None)
```
jamesbraza, line 255: Why do we need to mutate here, can't we just read?

I think doing a `deepcopy` on every LLM call is expensive

PR #280, opened 2025-04-27
Use router also for embeddings

tobiasdiez:
Currently, embeddings just call litellm and thus don't provide the same configuration possibilities as the other completion endpoints. To fix this, a router is now used similar to the other calls so that one can use eg the following for azure-hosted:
```python
embedding = "text-embedding-3-small"
embedding_config = {
    "model_list": [
        {
            "model_name": embedding,
            "litellm_params": {
                "model": "azure/text-embedding-3-small",  # azure/<your deployment name>
                "api_key": os.getenv("AZURE_API_KEY"),
                "api_version": "2024-04-01-preview",
                "api_base": os.getenv(
                    "AZURE_API_BASE"
                ),  # eg. https://openai-gpt-4-test-v-1.openai.azure.com/
            },
        },
    ]
}
```

### DIFF
--- a/packages/lmi/src/lmi/embeddings.py
+++ b/packages/lmi/src/lmi/embeddings.py
@@ -88,6 +88,19 @@ class LiteLLMEmbeddingModel(EmbeddingModel):
             " Router is not used here."
         ),
     )
+    _router: litellm.Router | None = None
+
+    @property
+    def router(self) -> litellm.Router:
+        if self._router is None:
+            router_kwargs: dict = self.config.get("router_kwargs", {})
+            if self.config.get("pass_through_router"):
+                self._router = PassThroughRouter(**router_kwargs)
+            else:
+                self._router = litellm.Router(
+                    model_list=self.config["model_list"], **router_kwargs
+                )
+        return self._router
 
     @model_validator(mode="before")
     @classmethod
@@ -146,7 +159,7 @@ class LiteLLMEmbeddingModel(EmbeddingModel):
                 )
             )
 
-            response = await track_costs(litellm.aembedding)(
+            response = await track_costs(self.router.aembedding)(
                 model=self.name,
                 input=texts[i : i + batch_size],
                 dimensions=self.ndim,
--- a/packages/lmi/src/lmi/embeddings.py
+++ b/packages/lmi/src/lmi/embeddings.py
@@ -95,10 +95,7 @@ class LiteLLMEmbeddingModel(EmbeddingModel):
     def router(self) -> litellm.Router:
         if self._router is None:
             router_kwargs: dict = self.config.get("router_kwargs", {})
-            if (
-                self.config.get("pass_through_router")
-                or "model_list" not in self.config
-            ):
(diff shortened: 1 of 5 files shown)


packages/lmi/src/lmi/embeddings.py
```diff
@@ -88,6 +88,19 @@ class LiteLLMEmbeddingModel(EmbeddingModel):
...
         ),
     )
+    _router: litellm.Router | None = None
+
+    @property
+    def router(self) -> litellm.Router:
+        if self._router is None:
+            router_kwargs: dict = self.config.get("router_kwargs", {})
+            if self.config.get("pass_through_router"):
+                self._router = PassThroughRouter(**router_kwargs)
```
jamesbraza, line 98: I think you need to import this -- can you get `pre-commit` and `lint` CI to pass?

jamesbraza (commented):
Nice work, I like the idea 👍

packages/lmi/src/lmi/embeddings.py
tobiasdiez: Hopefully the last commit did the job.

packages/lmi/src/lmi/embeddings.py
```diff
@@ -88,6 +89,20 @@ class LiteLLMEmbeddingModel(EmbeddingModel):
             " Router is not used here."
         ),
     )
+    _router: litellm.Router | None = None
+
+    @property
+    def router(self) -> litellm.Router:
+        if self._router is None:
+            router_kwargs: dict = self.config.get("router_kwargs", {})
+            if self.config.get("pass_through_router")
```
jamesbraza, line 98: This is a syntax error, need `:`

tobiasdiez: Should be fixed now.

jamesbraza: Thanks, but can you get `pre-commit` to pass too, it's still failing

packages/lmi/src/lmi/embeddings.py
jamesbraza: I just pushed a commit to fix `isort`

packages/lmi/src/lmi/embeddings.py
```diff
@@ -88,6 +88,19 @@ class LiteLLMEmbeddingModel(EmbeddingModel):
...
     )
+    _router: litellm.Router | None = None
+
+    @property
+    def router(self) -> litellm.Router:
+        if self._router is None:
+            router_kwargs: dict = self.config.get("router_kwargs", {})
+            if self.config.get("pass_through_router"):
+                self._router = PassThroughRouter(**router_kwargs)
+            else:
```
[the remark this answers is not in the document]
jamesbraza: Let's YAGNI on this nice-ification of a configuration issue. I am okay with a `KeyError`, it's direct and understandable

tobiasdiez:
Thanks for finishing this PR; I was pretty busy the last days.

PR #299, opened 2025-06-15
Helper function to defeat `litellm`'s `MAX_CALLBACKS` error

jamesbraza:
Adding a workaround for https://github.com/BerriAI/litellm/issues/9792


maykcaldas (approved):
Thanks for that! That has been annoying me for a while now

PR #307, opened 2025-07-01
Cleaned up the reasoning extraction logic

maykcaldas:
litellm fixed https://github.com/BerriAI/litellm/issues/8130 and estabilished an interface for a few other reasoning models than deepseek. This PR cleans the deepseek-specific code we had and follows `litellm` current interface


packages/lmi/tests/test_llms.py
```diff
@@ -862,6 +848,35 @@ async def test_deepseek_model(
             assert result.reasoning_content
             assert outputs[i] == result.text
 
+    @pytest.mark.vcr(match_on=[*VCR_DEFAULT_MATCH_ON, "body"])
+    @pytest.mark.asyncio
+    async def test_anthropic_model(self) -> None:
+        llm = LiteLLMModel(
+            name="anthropic/claude-3-7-sonnet-20250219",
```
jamesbraza: Can we use `CommonLLMNames.ANTHROPIC_TEST` here? And maybe we update its value

maykcaldas: I hardcorded 3.7-sonnet because the anthropic we use for test is 3.5-haiku, which doesn't support reasoning. And I didn't want to change everything to 3.7-sonnet

jamesbraza: Oh gotchu and sg.

Can you then:
- Add a code comment documenting this, something like `# Use 3.7 Sonnet since it supports reasoning`
- Add `CLAUDE_37_SONNET = "claude-3-7-sonnet-20250219"` to `CommonLLMNames` (no need to change `ANTHROPIC_TEST`)

maykcaldas: Done! Thanks for the suggestion!!

jamesbraza (approved):
Nice work!! Appreciate the tech debt burn down 🙏

PR #310, opened 2025-07-09
Add notes about vertex to README

whitead:
I keep forgetting how to do this, so added it to README

### DIFF
--- a/packages/lmi/README.md
+++ b/packages/lmi/README.md
@@ -391,7 +391,7 @@ Vertex requires a bit of extra set-up. First, install the extra dependency for a
 pip install google-api-python-client
 ```
 
-and then you need to configure which region/project your using for the model calls.
+and then you need to configure which region/project you're using for the model calls.
 Here's a complete example:
 
 ```py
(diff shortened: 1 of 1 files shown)


packages/lmi/README.md
```diff
@@ -383,6 +384,27 @@ result = await llm.call_single(
 # the calculator invocation with x=2, y=2, operation="+"
 ```
 
+### Vertex
+
+Vertex requires a bit of extra set-up. First, install the extra dependency for auth:
+
+```sh
+pip install google-api-python-client
```
jamesbraza, line 392: Why do we have to install this if your below code snippet doesn't use it?

jamesbraza: Mainly, if you're using this to auth up front, can you also document the command used with this Python package

whitead: Sure, you get a confusing import error if you don't run that first. But yea, will mention auth command too.

jamesbraza (approved):
Nice work, thanks

PR #314, opened 2025-07-10
Made it easier to wrap an action

whitead:
Introduced a new method (and added to existing agents) to simplify plain python `get_asv`


sidnarayanan (approved):
Nice, looks clean

PR #316, opened 2025-07-18
Enabled stream on lmi

maykcaldas:
This fixed our fake streaming mode we had with `acompletion_iter`. Now setting `stream` to `True` in both `call` and `call_single` returns an `AsyncGenerator` 

Also allowed string inputs in `call` for convenience

# TODO: Need to make sure other repos will not break because of the typing change. I typed overload to avoid that


packages/lmi/src/lmi/llms.py
```diff
@@ -245,22 +276,30 @@ async def call(  # noqa: C901, PLR0915
...
 
         Returns:
-            A list of LLMResult objects containing the result of the call.
+            A list of LLMResult objects containing the result of the call when stream=False,
+            or an AsyncGenerator[LLMResult] when stream=True.
 
         Raises:
             ValueError: If the LLM type is unknown.
         """
+        if isinstance(messages, str):
```
jamesbraza, line 289: If we want to do this here, can we remove it from `call_single`?

packages/lmi/src/lmi/llms.py
```diff
@@ -245,22 +276,30 @@ async def call(  # noqa: C901, PLR0915
             output_type: The type of the output.
             tools: A list of tools to use.
             tool_choice: The tool choice to use.
+            stream: Whether to stream the response or return all results at once.
             kwargs: Additional keyword arguments for the chat completion.
 
         Returns:
-            A list of LLMResult objects containing the result of the call.
+            A list of LLMResult objects containing the result of the call when stream=False,
+            or an AsyncGenerator[LLMResult] when stream=True.
```
jamesbraza: Two comments:

- Indent the later lines by 4
- Consider not restating the type hints/variable names, as that can lead to drift over time

For example:

```none
When not streaming, it's a list of result objects for each call, otherwise
    it's an async generator of result objects.
```

packages/lmi/src/lmi/llms.py
```diff
@@ -328,48 +367,83 @@ async def call(  # noqa: C901, PLR0915
...
+                    possibly_awaitable_result = self.llm_result_callback(result)
+                    if isawaitable(possibly_awaitable_result):
+                        await possibly_awaitable_result
+            return results
+
+        # If streaming, return an AsyncGenerator[LLMResult]
+        if tools:
+            raise NotImplementedError("Using tools with streaming is not supported")
+        if callbacks:
+            raise NotImplementedError("Using callbacks with streaming is not supported")
```
jamesbraza, line 399: Can you reword to say "not yet supported"

Also, can you add a comment to each of these saying why they're not supported?

packages/lmi/src/lmi/llms.py
```diff
@@ -328,48 +367,83 @@ async def call(  # noqa: C901, PLR0915
...
+        name: str | None = ...,
+        output_type: type[BaseModel] | TypeAdapter | JSONSchema | None = ...,
+        tools: list[Tool] | None = ...,
+        tool_choice: Tool | str | None = ...,
+        stream: bool = False,
+        **kwargs,
+    ) -> LLMResult: ...
+
+    @overload
+    async def call_single(  # type: ignore[overload-cannot-match]
```
jamesbraza: This type ignore -- why do you have it? We should have no type ignores here imo, otherwise it means the typing is wrong

PR #324, opened 2025-08-18
Move out dm-tree and networkx from main dependencies

whitead:
I've made dm-tree, networkx, and usearch optional. This will help paper-qa reduce dependencies when it needs ldp. 

Introduced new subpackage `ops` that will include those two, which is required for gradients/backwards pass on compute graphs.

### DIFF
--- a/src/ldp/data_structures.py
+++ b/src/ldp/data_structures.py
@@ -17,7 +17,6 @@ from ldp.graph.op_utils import _lazy_import_networkx
 from ldp.utils import discounted_returns
 
 if TYPE_CHECKING:
-    import networkx as nx  # noqa: F401
     from aviary.core import Environment
 
 
--- a/src/ldp/graph/op_utils.py
+++ b/src/ldp/graph/op_utils.py
@@ -1,18 +1,14 @@
 import contextvars
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
-from typing import TYPE_CHECKING
+from types import ModuleType
 from uuid import UUID, uuid4
 
 from aviary.core import is_coroutine_callable
 from pydantic import BaseModel, field_serializer, field_validator
 
-if TYPE_CHECKING:
-    import networkx as nx
-    import tree
 
-
-def _lazy_import_networkx() -> "nx":
+def _lazy_import_networkx() -> ModuleType:
     try:
         import networkx as nx
     except ImportError as e:
@@ -23,7 +19,7 @@ def _lazy_import_networkx() -> "nx":
     return nx
 
 
-def _lazy_import_tree() -> "tree":
+def _lazy_import_tree() -> ModuleType:
     try:
         import tree
     except ImportError as e:


src/ldp/graph/op_utils.py
```diff
@@ -7,6 +7,30 @@
...
+        import networkx as nx
+    except ImportError as e:
+        raise ImportError(
+            "networkx is required for compute graph operations. "
+            "Please install it with: pip install ldp[ops]"
+        ) from e
+    return nx
+
+
+def _lazy_import_tree():
```
jamesbraza: Can you type hint these to return `types.ModuleType`

pyproject.toml
```diff
@@ -76,11 +74,17 @@ nn = [
     "dask-cuda>=24.8.2",
     "dask-jobqueue",
     "dask[distributed]",
+    "dm-tree",
+    "networkx[default]~=3.4",  # Pin for pydot fix
```
jamesbraza:
```suggestion
    "ldp[ops]",
```

pyproject.toml
```diff
@@ -76,11 +74,17 @@ nn = [
...
     "tokenizers>0.20",
     "torch>=2.5,<2.6",  # Temporarily pin <2.6 until someone fixes our CI with newer versions
     "transformers>=4.46,<4.51",  # Temporarily pin <4.51 until someone fixes our CI with newer versions, and also https://github.com/huggingface/transformers/issues/37339
     "wandb",
 ]
+ops = [
```
jamesbraza: Can you document this in the `REAMDE.md` somewhere, just a quick sentence

pyproject.toml
```diff
@@ -23,18 +23,15 @@ classifiers = [
 ]
 dependencies = [
     "aiofiles",
-    "dm-tree",
```
jamesbraza, line 26: I agree with `networkx` lazy import. Though without `dm-tree`, graph won't work, which means agents won't work, which means runners and nn won't work... so basically the whole package doesn't work.

Even though it's nice to move `dm-tree` to an extra, idk if it actually makes sense. Perhaps PaperQA should be tasteful in importing LDP and bypass this

whitead: Funnily enough - rollout crashes on networkx imports, not dm-tree.

src/ldp/data_structures.py
```diff
@@ -123,11 +124,11 @@ def done(self) -> bool:
             return False
         return self.steps[-1].done
 
+    # async for backwards compatibility
     async def to_jsonl(self, filename: str | os.PathLike) -> None:
-        async with aiofiles.open(filename, "w") as f:
```
jamesbraza: Why remove `aiofiles` here? We still have it as a primary dep

src/ldp/graph/op_utils.py
```diff
@@ -1,11 +1,40 @@
...
 
 from aviary.core import is_coroutine_callable
 from pydantic import BaseModel, field_serializer, field_validator
 
+if TYPE_CHECKING:
+    import networkx as nx
+    import tree
+
+
+def _lazy_import_networkx() -> "nx":
```
jamesbraza: So in Python type hints are types, not instances.

Are you sure this type hint as a module instance works?

src/ldp/data_structures.py
whitead: reverted this - nice catch

src/ldp/graph/op_utils.py
jamesbraza: Okay I looked into this. On the below script, `mypy==1.17.1` will say:

```python
from typing import TYPE_CHECKING, reveal_type

if TYPE_CHECKING:
    import math

    import networkx as nx


def import_math() -> "math":
    import math

    return math


def import_nx() -> "nx":
    import networkx as nx

    return nx
```

```none
a.py: note: In function "import_math":
a.py:9:22: error: Module "math" is not valid as a type  [valid-type]
    def import_math() -> "math":
                         ^
```

So `mypy` is not warning us about `nx` but it does for `math`. The reason is that we disabled `mypy` on `networkx`: https://github.com/Future-House/ldp/blob/v0.31.0/pyproject.toml#L190

`networkx` and `tree` are both untyped, so that's why `mypy` wasn't complaining.

So we do need `types.ModuleType` for the return type

src/ldp/data_structures.py
```diff
@@ -9,16 +9,18 @@
...
-import networkx as nx
 from aviary.core import Message, ToolRequestMessage, ToolResponseMessage, join
 from pydantic import BaseModel, ConfigDict, Field, JsonValue, field_validator
 
 from ldp.graph import OpResult
+from ldp.graph.op_utils import _lazy_import_networkx
 from ldp.utils import discounted_returns
 
 if TYPE_CHECKING:
+    import networkx as nx  # noqa: F401
```
jamesbraza:
```suggestion
```

We don't need this one

src/ldp/graph/op_utils.py
whitead: image

jamesbraza (approved):
Nice work here

PR #327, opened 2025-08-20
Change traverse default behavior to avoid nx import

whitead:
@sidnarayanan really need you on this one - don't know the consequences of this change in default behavior.

### DIFF
--- a/src/ldp/graph/ops.py
+++ b/src/ldp/graph/ops.py
@@ -246,7 +246,7 @@ class OpResult(Generic[TOutput_co]):
 
     def traverse(
         self,
-        topological_order: bool = True,
+        topological_order: bool = False,
         filter_fn: Callable[[OpResult], bool] = lambda _: True,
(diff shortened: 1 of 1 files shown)


src/ldp/graph/ops.py
```diff
@@ -246,7 +246,7 @@ def get_upstream_results(self, op: str | Op) -> Iterator[OpResult]:
 
     def traverse(
         self,
-        topological_order: bool = True,
+        topological_order: bool = False,
```
sidnarayanan, line 249: I'm cool with changing the default, but can you pass `topological_order=True` here: https://github.com/Future-House/ldp/blob/3d874c59088f70efbe323878bd52b748438bcb50/src/ldp/graph/ops.py#L119 ? That's the only place we currently require topological sorting.

whitead: @copilot address this comment

PR #328, opened 2025-08-21
Handle top logprobs in litellm call

albertbou92:


packages/lmi/tests/test_llms.py
```diff
@@ -122,8 +135,24 @@ async def test_call(self, config: dict[str, Any]) -> None:
         if llm.config["model_list"][0]["litellm_params"].get("logprobs"):
             assert isinstance(result.logprob, float)
             assert result.logprob <= 0
+            # Test top_logprobs only for OpenAI models (top_logprobs is OpenAI-specific)
+            is_openai_model = "gpt" in llm.name.lower() or "openai" in llm.name.lower()
```
jamesbraza: Can you use `litellm.get_llm_provider` here

packages/lmi/src/lmi/types.py
```diff
@@ -79,6 +79,9 @@ class LLMResult(BaseModel):
     logprob: float | None = Field(
         default=None, description="Sum of logprobs in the completion."
     )
+    top_logprobs: list[list[tuple[str, float]]] = Field(
+        default_factory=list, description="Top logprobs for each position in the completion."
+    )
```
jamesbraza: I would rather see `list[list[tuple[str, float]]] | None` with `default=None` to avoid injecting information such as empty list

packages/lmi/tests/test_llms.py
maykcaldas: what about o-series?

packages/lmi/src/lmi/types.py
```diff
@@ -79,6 +79,9 @@ class LLMResult(BaseModel):
     logprob: float | None = Field(
         default=None, description="Sum of logprobs in the completion."
     )
+    top_logprobs: list[list[tuple[str, float]]] = Field(
```
maykcaldas: Can this be `list[list[tuple[str, float]]] | None` and default to None? 

Then have `extract_top_logprobs` return None if the LLM doesn't return a logprob. I'm thinking about compatibility with providers other than openai. 

And then in the tests, you don't need to check if `is_openai`. Instead, you check if `top_logprobs` is a float and then if there is a `top_logprobs` array being returned

packages/lmi/tests/test_llms.py
maykcaldas: Added a comment above about that. I think `LLMResult` should just have `LLMResult.top_logprobs` set to `None` if it can't be computed

albertbou92: Nice!! I did not know this method existed :)

packages/lmi/src/lmi/types.py
albertbou92: makes sense

packages/lmi/tests/test_llms.py
albertbou92: makes sense, added

packages/lmi/src/lmi/types.py
albertbou92: The issues is that litellm crashes if these params are provided for a non openai model

packages/lmi/pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
     "pydantic~=2.0,>=2.10.1",
     "tiktoken>=0.4.0",
     "typing-extensions; python_version <= '3.11'",  # for typing.override
+    "openai>=1,<1.100.0",  # Pin to keep recent, downpin for https://github.com/BerriAI/litellm/issues/13710
```
jamesbraza: I don't think we need this any more, `openai` fixed themselves

packages/lmi/src/lmi/llms.py
```diff
@@ -118,6 +121,22 @@ def sum_logprobs(choice: litellm.utils.Choices | list[float]) -> float | None:
...
+    
+    content = getattr(completion.logprobs, "content", None)
+    if not content or not isinstance(content, list):
+        return None
+    
+    out = []
+    for pos in content:
+        tops = getattr(pos, "top_logprobs", []) or []
+        out.append([(t.token, float(t.logprob)) for t in tops])
+    return out
```
jamesbraza: Can we make this a list comprehension?

packages/lmi/tests/test_llms.py
```diff
@@ -58,6 +58,18 @@ def test_instantiation_methods(self) -> None:
         assert model4.config["model_list"][0]["litellm_params"]["temperature"] == 0.5
         assert model4.config["model_list"][0]["litellm_params"]["max_tokens"] == 100
 
+        # Test logprobs and top_logprobs configuration passing through (OpenAI-specific)
+        name = CommonLLMNames.OPENAI_TEST.value
+        config_with_logprobs = {
+            "logprobs": True, 
+            "top_logprobs": 20,
```
jamesbraza: There's trailing whitespace here, why is that?

packages/lmi/pyproject.toml
jamesbraza: I think you thumbs up'd this and resolved it but I still see it. This PR shouldn't need to change `pyproject.toml` or `uv.lock`, we'd only make this change when updating deps

packages/lmi/tests/test_llms.py
albertbou92: where?

jamesbraza: screenshot of trailing whitespace

packages/lmi/pyproject.toml
albertbou92: yeah sorry. I removed it but then I asked @maykcaldas to push some code over mine. So I had to remove it again.

packages/lmi/tests/test_llms.py
albertbou92: got it. good catch, thanks. Its gone now.

.python-version
```diff
@@ -1 +1 @@
-3.13
+3.11
```
jamesbraza: Oh can we not do this? Why do we need to do this here?

jamesbraza: Can we at least do 3.12 🤓

packages/lmi/tests/cassettes/TestLiteLLMEmbeddingModel.test_caching.yaml
```diff
@@ -1,215 +0,0 @@
-interactions:
```
jamesbraza: Can we bring these cassettes back? We need them, what's happening now is CI itself just regens the cassettes

Also, why are embedding cassettes being changed, they're unrelated to `top_logprobs`

.python-version
jamesbraza: https://github.com/Future-House/ldp/pull/330

packages/lmi/tests/cassettes/TestLiteLLMEmbeddingModel.test_caching.yaml
jamesbraza: https://github.com/Future-House/ldp/pull/330

PR #332, opened 2025-09-09
pin torch 2.6

albertbou92:


pyproject.toml
```diff
@@ -76,8 +76,8 @@ nn = [
     "dask[distributed]",
     "ldp[scg]",
     "tokenizers>0.20",
-    "torch>=2.5,<2.6",  # Temporarily pin <2.6 until someone fixes our CI with newer versions
-    "transformers>=4.46,<4.51",  # Temporarily pin <4.51 until someone fixes our CI with newer versions, and also https://github.com/huggingface/transformers/issues/37339
+    "torch==2.6",
+    "transformers",
```
jamesbraza:
```suggestion
    "transformers>=4.51",  # Pin for py.typed
```

src/ldp/nn/handlers/chunking.py
```diff
@@ -113,37 +116,51 @@ def dechunkify(
             return torch.cat(real_outputs, dim=0)
 
         if isinstance(real_outputs[0], GenerateDecoderOnlyOutput):
-            sequences = []
-            for output in real_outputs:
-                sequences.extend(output.sequences)
```
jamesbraza, line 118: Lol at this

src/ldp/nn/handlers/chunking.py
```diff
@@ -113,37 +116,51 @@ def dechunkify(
...
-                        if i < len(output.scores):
+                        if output.scores and i < len(output.scores):
                             scores_step_i.extend(list(output.scores[i]))
                         else:
-                            # Add bsz Nones for this worker, since it did not reach token `i`.
-                            scores_step_i.extend([None] * output.sequences.shape[0])
-
-                    scores.append(scores_step_i)
-
+                            # Add dummy values for this worker, since it did not reach token `i`.
```
jamesbraza: Do you mind just stating what values are the dummy ones? For example:

> Add dummy values X and Y for this worker, since it did not reach token `i`.

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -899,24 +936,25 @@ def decollate_fn_transformer_decoder(
...
+                    score[i][None, :].to(dtype=torch.float32)
                     for score in batched_output.scores
                     if (score[i] is not None)
-                ]
-                if batched_output.scores
-                else None
+                ],
             ),
-            # There are other fields in the batched output that we can add here,
-            # but no calling code is using them, so ignore for now.
```
jamesbraza, line 916: Why clobber this comment?

tests/test_nn_models.py
```diff
@@ -93,9 +92,12 @@ def test_dechunkify(self):
 
         # Simulate outputs from handlers
         outputs_list = []
+        from typing import cast
```
jamesbraza: Can you do this at the module level

src/ldp/nn/handlers/transformer_handler.py
albertbou92: re-adding

src/ldp/nn/graph/llm_call_op.py
```diff
@@ -87,13 +88,15 @@ def prep_messages_for_tokenizer(xi: Messages) -> list[dict]:
...
+    def prep_tools_for_tokenizer(
+        tools: list[Tool] | None,
+    ) -> list[dict[Any, Any] | Callable[..., Any]] | None:
         """Prepare tools for the tokenizer by transforming them into a JSON schema."""
         if not tools:
             return None
 
         # TODO: should be able to switch to tool.info.model_dump() here
-        tools_list = []
+        tools_list: list[dict[Any, Any] | Callable[..., Any]] = []
```
sidnarayanan: Why is tool allowed to be a callable? I only see dicts being added.

src/ldp/nn/handlers/chunking.py
sidnarayanan: Is this correct? `output.sequences` is a 2D tensor of shape `(bsz, seq_len)`.  The old code flattened this into a list of 1D tensors. The new code makes a list of 2D tensors.

src/ldp/nn/handlers/chunking.py
```diff
@@ -113,37 +116,52 @@ def dechunkify(
...
+            ).to(torch.long)
+            scores_tuple = (
+                cast(tuple[torch.FloatTensor], tuple(scores)) if scores else None
+            )
             return GenerateDecoderOnlyOutput(
-                sequences=sequences,
-                scores=scores,
+                sequences=cast(torch.LongTensor, padded_sequences),
+                scores=scores_tuple,
             )
```
sidnarayanan: Discussed in person - this is a big change and won't be folded into this PR. This change returns different tensors than before, and we need experiments to understand downstream consequences.

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -186,7 +195,7 @@ class AsyncTransformerInterface(ModuleExecutionInterface, AsyncTorchModule, ABC)
     async def __call__(  # type: ignore[override]
         self,
         inputs: str | BatchEncoding | list[dict],
-        tools_json: list[dict] | None = None,
+        tools_json: list[dict[Any, Any] | Callable[..., Any]] | None = None,
```
sidnarayanan: Same question as above - when are we passing callables? also `dict` is `dict[Any, Any]`.

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -201,7 +210,10 @@ def model_generate(model: PreTrainedModel, *args, **kwargs):
                 f"model.generate() input_ids shape: {kwargs['input_ids'].shape}, rank"
                 f" {os.environ.get('RANK')}"
             )
-            return model.generate(
+            generate_fn = getattr(model, "generate", None)
+            if not callable(generate_fn):
+                raise TypeError("Model does not implement generate()")
+            return generate_fn(
```
sidnarayanan:
```suggestion
           if not isinstance(model, GenerationMixin):
                raise TypeError("model_generate only supports models that inherit from GenerationMixin")
            return model.generate(
```

This is both stricter (guarantees the signature of `generate` too) and easier to read.

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -220,13 +232,19 @@ def __init__(self, config: TransformerHandlerConfig):
 
         match config.lm_type:
             case LMType.GENERATION:
-                tokenizer, model = config.lm_config.get_causal_lm()
+                tokenizer, model = cast(
+                    tuple[PreTrainedTokenizer, PreTrainedModel],
+                    config.lm_config.get_causal_lm(),
+                )
```
sidnarayanan: Perhaps better to change the typing of `get_causal_lm` instead of having to cast everywhere?

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -420,9 +438,9 @@ def _exec_func(
         kwargs = tree.map_structure(to_device, kwargs)
 
         try:
-            with torch.autocast(
-                device_type=self.module.device.type, dtype=self.module.dtype
-            ):
+            device_type: str = str(self.module.device.type)
+            dtype: torch.dtype | None = getattr(self.module, "dtype", None)
```
sidnarayanan: We should blow up if `dtype` is missing - it's assumed to be there.

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -610,13 +628,32 @@ async def _batched_call(self, batch_kwargs: dict[str, Any]):
...
+    @overload
+    def _submit_and_gather(
+        self,
+        func: Callable[Concatenate[ParallelTransformerHandler, TParams], TOutputType]
+        | str,
+        *args,
+        split_data: Literal[True],
+        **kwargs,
+    ) -> TOutputType: ...
+
```
sidnarayanan, line 636: Do we have to introduce a new `TOutputType` here or can we reuse `TReturn`?

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -899,24 +936,27 @@ def decollate_fn_transformer_decoder(
...
+        GenerateDecoderOnlyOutput(
+            sequences=cast(
+                torch.LongTensor,
+                batched_output.sequences[i][None, :].to(dtype=torch.long),
+            ),
+            scores=cast(
+                tuple[torch.FloatTensor],
                 [
-                    score[i][None, :]
+                    score[i][None, :].to(dtype=torch.float32)
```
sidnarayanan: As above, let's not mess with logic to satisfy typing. I'd want to check this doesn't cause issues with mixed precision training before merging. If `GenerateDecoderOnlyOutput` doesn't like the dtypes we are passing, `type: ignore` it.

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -899,24 +936,27 @@ def decollate_fn_transformer_decoder(
...
-    return [
-        GenerateDecoderOnlyOutput({
-            "sequences": batched_output.sequences[i][None, :],
-            "scores": (
+    assert batched_output.scores is not None
+    outputs: list[GenerateDecoderOnlyOutput] = [
+        GenerateDecoderOnlyOutput(
+            sequences=cast(
+                torch.LongTensor,
+                batched_output.sequences[i][None, :].to(dtype=torch.long),
```
sidnarayanan: Same as below.

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -899,24 +936,27 @@ def decollate_fn_transformer_decoder(
...
+            scores=cast(
+                tuple[torch.FloatTensor],
                 [
-                    score[i][None, :]
+                    score[i][None, :].to(dtype=torch.float32)
                     for score in batched_output.scores
                     if (score[i] is not None)
-                ]
-                if batched_output.scores
-                else None
```
sidnarayanan, line 913: Also a logic change. Do we understand the consequences of not allowing missing scores? 

One issue I can think of: this will break if someone calls `model.generate` with default arguments, which do not return scores: https://huggingface.co/docs/transformers/v4.56.1/en/main_classes/text_generation#transformers.GenerationConfig.output_scores - remember, we did this in `molr1` to save memory during sampling.

src/ldp/nn/handlers/transformer_handler.py
```diff
@@ -977,13 +1017,18 @@ def _get_tokenized_inputs(
...
             inputs,
             tools=tools_json,
             return_tensors="pt",
             return_dict=True,
             add_generation_prompt=True,
         )
+        if not isinstance(result, BatchEncoding):
+            raise TypeError(
+                f"Expected BatchEncoding from apply_chat_template, got {type(result)}"
+            )
```
sidnarayanan: cast is sufficient here - `return_dict` guarantees a `BatchEncoding`

src/ldp/nn/graph/llm_call_op.py
albertbou92: This is the typing allowed by `tokenizer.apply_chat_template`.

Defining it as subset  (`list[dict[Any, Any]]`) is not allowed. Makes no sense to me.

In any case, we are just saying that it could be a callable if we ever want.

src/ldp/nn/handlers/transformer_handler.py
albertbou92: `PreTrainedModel` and `GenerationMixin` have incompatible method signatures, says mypy

src/ldp/nn/handlers/transformer_handler.py
albertbou92: This is the typing allowed by `tokenizer.apply_chat_template`.

Defining it as subset (`list[dict[Any, Any]]`) is not allowed. Makes no sense to me.

In any case, we are just saying that it could be a callable if we ever want.

src/ldp/nn/handlers/transformer_handler.py
albertbou92: I don't get it. Why should we blow up? why not falling to default dtype?

src/ldp/nn/handlers/transformer_handler.py
albertbou92: `dechunkify` returns `TOutputType`

`TOutputType = TypeVar("TOutputType", torch.Tensor, GenerateDecoderOnlyOutput)
`
Maybe it should returns a `TReturn`?

src/ldp/nn/graph/llm_call_op.py
sidnarayanan: I think it'd be preferable to make this `list[dict[Any, Any]]` and then `type: ignore` the pass to `apply_chat_template`.

They should've used `Sequence[...]` instead of `list` to avoid this issue. Not sure, but that might work on our end too.

src/ldp/nn/handlers/transformer_handler.py
sidnarayanan: Ok then I think this makes sense - `TOutputType` is more constrained than `TReturn` so it makes sense to use different type vars.

src/ldp/nn/handlers/transformer_handler.py
sidnarayanan: It's a logic error if we are missing `dtype`. Better to fail (as we were doing before) than falling back silently.

src/ldp/nn/handlers/transformer_handler.py
sidnarayanan: What's the mypy error?

src/ldp/nn/handlers/transformer_handler.py
albertbou92: Should we just ignore then? there is no point in casting either if we just dont know the type

sidnarayanan: Yep

src/ldp/nn/handlers/transformer_handler.py
albertbou92:
```
src/ldp/nn/handlers/transformer_handler.py: note: In member "model_generate" of class "AsyncTransformerInterface":
src/ldp/nn/handlers/transformer_handler.py:213:27: error: Subclass of
"PreTrainedModel" and "GenerationMixin" cannot exist: would have incompatible
method signatures  [unreachable]
                if isinstance(model, GenerationMixin):
```

sidnarayanan: https://github.com/huggingface/transformers/blob/cf084f5b40e19b5a5f946cee75bead6d4247b071/src/transformers/models/llama/modeling_llama.py#L413

`LlamaPreTrainedModel` inherits from `PreTrainedModel`.

`transformers`'s typing is imperfect - on the balance of satisfying mypy vs having cleaner/more functional code, let's pick the latter.

src/ldp/nn/handlers/chunking.py
albertbou92: yes, rolling back to previous code

src/ldp/nn/handlers/transformer_handler.py
albertbou92: happy to change, but is this that bad? maybe just a bit overkill 

```
generate_fn = getattr(model, "generate", None)
if not callable(generate_fn):
    raise TypeError("Model does not implement generate()")
```

src/ldp/nn/handlers/chunking.py
albertbou92: rolling back to previous code

src/ldp/nn/handlers/transformer_handler.py
sidnarayanan: Not a huge difference, but I figure if we are doing all this work to satisfy typing, we may as well get the call signature provided by `GenerationMixin`.

src/ldp/nn/handlers/transformer_handler.py
albertbou92: re-added

PR #335, opened 2025-09-12
Make the callbacks in cost_tracker async

AndyJCai:
Making the callbacks in cost_tracker.py asynchronous: job-event calls run after every LLM request, so we schedule them on the event loop to avoid blocking the main execution path and adding latency.

### DIFF
--- a/packages/lmi/src/lmi/cost_tracker.py
+++ b/packages/lmi/src/lmi/cost_tracker.py
@@ -1,4 +1,3 @@
-import asyncio
 import contextvars
 import logging
 from collections.abc import Awaitable, Callable
@@ -7,6 +6,7 @@ from functools import wraps
 from typing import Any, ParamSpec, TypeVar
 
 import litellm
+from typing_extensions import deprecated
 
 from lmi.types import LLMResponse
 
@@ -32,6 +32,8 @@ class CostTracker:
 
         This method only handles cost tracking and reporting. It does not execute
         callbacks, making it safe to use in any context without race conditions.
+
+        Note that this method is not async-friendly. Use `arecord` instead if you need to execute callbacks.
         """
         self.lifetime_cost_usd += litellm.cost_calculator.completion_cost(
             completion_response=response
@@ -41,7 +43,7 @@ class CostTracker:
             logger.info(f"Cumulative lmi API call cost: ${self.lifetime_cost_usd:.8f}")
             self.last_report = self.lifetime_cost_usd
 
-    async def record_with_callbacks(self, response: LLMResponse) -> None:
+    async def arecord(self, response: LLMResponse) -> None:
         """Record cost and execute all registered callbacks.
 
         This method handles both cost tracking and callback execution. Use this
@@ -114,7 +116,7 @@ def track_costs(
     async def wrapped_func(*args, **kwargs):
         response = await func(*args, **kwargs)
         if GLOBAL_COST_TRACKER.enabled.get():
-            await GLOBAL_COST_TRACKER.record_with_callbacks(response)
+            await GLOBAL_COST_TRACKER.arecord(response)
         return response
 
     return wrapped_func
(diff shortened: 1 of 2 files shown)


packages/lmi/src/lmi/cost_tracker.py
```diff
@@ -6,6 +6,7 @@
 from typing import Any, ParamSpec, TypeVar
 
 import litellm
+from typing_extensions import deprecated
```
jamesbraza: No need for this `deprecated`, we are zero-ver. Just remove it

packages/lmi/src/lmi/cost_tracker.py
```diff
@@ -20,15 +21,20 @@ def __init__(self):
         self.enabled = contextvars.ContextVar[bool]("track_costs", default=False)
         # Not a contextvar because I can't imagine a scenario where you'd want more fine-grained control
         self.report_every_usd = 1.0
-        self._callbacks: list[Callable[[LLMResponse], Any]] = []
+        # Support async callbacks only
+        self._callbacks: list[Callable[[LLMResponse], Awaitable[Any]]] = []
```
jamesbraza:
```suggestion
        self._callbacks: list[Callable[[LLMResponse], Awaitable]] = []
```

[style nit] don't put `Any`, just let it be inferred

packages/lmi/src/lmi/cost_tracker.py
```diff
@@ -20,15 +21,20 @@ def __init__(self):
...
 
-    def add_callback(self, callback: Callable[[LLMResponse], Any]) -> None:
+    def add_callback(self, callback: Callable[[LLMResponse], Awaitable[Any]]) -> None:
         self._callbacks.append(callback)
 
-    def record(
-        self,
-        response: LLMResponse,
-    ) -> None:
+    def record(self, response: LLMResponse) -> None:
```
jamesbraza:
```suggestion
    async def record(self, response: LLMResponse) -> None:
```

Let's just do the full-blown move

packages/lmi/src/lmi/cost_tracker.py
```diff
@@ -20,15 +21,20 @@ def __init__(self):
...
+        """Record cost without executing callbacks.
+
+        This method only handles cost tracking and reporting. It does not execute
+        callbacks, making it safe to use in any context without race conditions.
```
jamesbraza: > making it safe to use in any context without race conditions

It's implied all of our code is safe without race conditions, no need to document this. Let's clean up AI superfluous comments

packages/lmi/src/lmi/cost_tracker.py
```diff
@@ -20,15 +20,13 @@ def __init__(self):
         self.enabled = contextvars.ContextVar[bool]("track_costs", default=False)
         # Not a contextvar because I can't imagine a scenario where you'd want more fine-grained control
         self.report_every_usd = 1.0
-        self._callbacks: list[Callable[[LLMResponse], Any]] = []
+        # Support async callbacks only
```
jamesbraza:
```suggestion
```

Don't put comments that directly mirror the code. Either add "why" to your comment, but imo this should just be removed

We didn't have a comment earlier saying "Support sync callbacks only" either

packages/lmi/tests/test_cost_tracking.py
```diff
@@ -212,10 +195,60 @@ async def test_multiple_callbacks_with_one_failing(self, caplog):
...
+            assert GLOBAL_COST_TRACKER.lifetime_cost_usd > 0
+
+    @pytest.mark.asyncio
+    async def test_record_method_with_callbacks(self):
+        """Test that the record method handles both cost tracking and callbacks."""
+        mock_response = MagicMock(cost=0.01)
+        callback_calls: list[Any] = []
+
+        async def async_callback(response):
+            await asyncio.sleep(0.1)
```
jamesbraza:
```suggestion
```

Why do we have `await asyncio.sleep(0.1)`? We don't need this in the test

packages/lmi/tests/test_cost_tracking.py
```diff
@@ -212,10 +195,60 @@ async def test_multiple_callbacks_with_one_failing(self, caplog):
...
+        ):
+            result = await anext(wrapper)
+
+            assert result == mock_response
+            assert len(callback_calls) == 1
+            assert callback_calls[0] == mock_response
+            assert GLOBAL_COST_TRACKER.lifetime_cost_usd > 0
+
+    @pytest.mark.asyncio
+    async def test_record_method_with_callbacks(self):
```
jamesbraza:
```suggestion
    async def test_callback_succeeds(self):
```

Also, can you move this up to the top of `TestCostTrackerCallback` to preserve the diff?

packages/lmi/tests/test_cost_tracking.py
AndyJCai: I believe I added it because lint was complaining that the async function has no async statements it. I can deprecate the lint instead

packages/lmi/tests/test_cost_tracking.py
```diff
@@ -212,10 +212,36 @@ async def test_multiple_callbacks_with_one_failing(self, caplog):
...
+
+    @pytest.mark.asyncio
+    async def test_async_context_with_stream_wrapper(self):
+        mock_stream = MagicMock()
+        mock_response = MagicMock(cost=0.01)
+        mock_stream.__anext__ = AsyncMock(return_value=mock_response)
+
+        wrapper = TrackedStreamWrapper(mock_stream)
+
+        callback_calls: list[Any] = []
```
jamesbraza:
```suggestion
        callback_calls = []
```

Doing `list[Any]` defeats type inference, effectively dumbing `mypy`. Let's let our tests be typed, so `mypy` can check here too

jamesbraza (approved):

PR #336, opened 2025-09-16
Add support for a messages sliding window

albertbou92:
Some environments might not need to keep track of all the previous messages history (e.g. envs that keep a plan in the env state and keep updating it).

This PR allows agents to define a sliding window to only keep in context the last N chat transitions.

### DIFF
--- a/src/ldp/agent/react_agent.py
+++ b/src/ldp/agent/react_agent.py
@@ -117,6 +117,16 @@ class ReActAgent(BaseModel, Agent[SimpleAgentState]):
         description="See SimpleAgentState.hide_old_env_states.",
     )
 
+    hide_old_action_content: bool = Field(
+        default=False,
+        description="See SimpleAgentState.hide_old_action_content.",
+    )
+
+    sliding_window: int = Field(
+        default=0,
+        description="Number of previous trajectory transitions to keep in the state. 0 means all previous transitions.",
+    )
+
     @classmethod
     def make_act_agent(cls, **kwargs) -> Self:
         single_prompt = kwargs.pop("single_prompt", False)
@@ -151,7 +161,10 @@ class ReActAgent(BaseModel, Agent[SimpleAgentState]):
 
     async def init_state(self, tools: list[Tool]) -> SimpleAgentState:
         return SimpleAgentState(
-            tools=tools, hide_old_env_states=self.hide_old_env_states
+            tools=tools,
+            hide_old_env_states=self.hide_old_env_states,
+            hide_old_action_content=self.hide_old_action_content,
+            sliding_window=self.sliding_window,
         )
 
     @staticmethod
--- a/src/ldp/agent/react_agent.py
+++ b/src/ldp/agent/react_agent.py
@@ -122,8 +122,8 @@ class ReActAgent(BaseModel, Agent[SimpleAgentState]):
         description="See SimpleAgentState.hide_old_action_content.",
     )
 
-    sliding_window: int = Field(
-        default=-1,
+    sliding_window: int | None = Field(
+        default=None,
         description=SimpleAgentState.model_fields["sliding_window"].description,
     )
 
--- a/src/ldp/agent/react_agent.py
+++ b/src/ldp/agent/react_agent.py
@@ -124,7 +124,7 @@ class ReActAgent(BaseModel, Agent[SimpleAgentState]):
 
     sliding_window: int | None = Field(
         default=None,
-        description=SimpleAgentState.model_fields["sliding_window"].description,
+        description="See SimpleAgentState.sliding_window.",
     )
 
     @classmethod
--- a/src/ldp/agent/react_agent.py
+++ b/src/ldp/agent/react_agent.py
@@ -123,8 +123,8 @@ class ReActAgent(BaseModel, Agent[SimpleAgentState]):
     )
 
     sliding_window: int = Field(
-        default=0,
-        description="Number of previous trajectory transitions to keep in the state. 0 means all previous transitions.",
+        default=-1,
+        description=SimpleAgentState.model_fields["sliding_window"].description,
     )
 
     @classmethod
--- a/src/ldp/agent/simple_agent.py
+++ b/src/ldp/agent/simple_agent.py
@@ -10,6 +10,7 @@ from pydantic import BaseModel, ConfigDict, Field
 
 from ldp.graph import ConfigOp, FxnOp, LLMCallOp, OpResult, compute_graph
 from ldp.llms import prepend_sys
+from ldp.utils import split_message_transitions
 
 from . import DEFAULT_LLM_COMPLETION_TIMEOUT
 from .agent import Agent
@@ -17,6 +18,10 @@ from .agent import Agent
 
 class HiddenEnvStateMessage(EnvStateMessage):
     content: str = "[Previous environment state - hidden]"
+
+
+class HiddenTransitionsMessage(Message):
+    content: str = "[Previous transitions - hidden]"
 
 
 def hide_action_content(msg: ToolRequestMessage) -> ToolRequestMessage:
@@ -39,6 +44,10 @@ class SimpleAgentState(BaseModel):
     hide_old_action_content: bool = Field(
         default=False,
         description="If True, will hide the content of old ToolRequestMessages.",
+    )
+    sliding_window: int = Field(
+        default=0,
+        description="Number of previous trajectory transitions to keep in the state. 0 means all previous transitions.",
     )
 
     def get_next_state(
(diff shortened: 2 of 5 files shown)


src/ldp/agent/react_agent.py
```diff
@@ -117,6 +117,16 @@ class ReActAgent(BaseModel, Agent[SimpleAgentState]):
...
     )
 
+    hide_old_action_content: bool = Field(
+        default=False,
+        description="See SimpleAgentState.hide_old_action_content.",
+    )
+
+    sliding_window: int = Field(
+        default=0,
+        description="Number of previous trajectory transitions to keep in the state. 0 means all previous transitions.",
```
jamesbraza: Can we use `SimpleAgentState.model_fields["sliding_window"].description`/`default` here

src/ldp/agent/simple_agent.py
```diff
@@ -19,6 +21,10 @@ class HiddenEnvStateMessage(EnvStateMessage):
     content: str = "[Previous environment state - hidden]"
 
 
+class HiddenTransitionsMessage(Message):
```
jamesbraza: Do you mind using `Message.info` over introducing more classes

src/ldp/agent/simple_agent.py
```diff
@@ -40,6 +46,10 @@ class SimpleAgentState(BaseModel):
         default=False,
         description="If True, will hide the content of old ToolRequestMessages.",
     )
+    sliding_window: int = Field(
+        default=0,
+        description="Number of previous trajectory transitions to keep in the state. 0 means all previous transitions.",
```
jamesbraza: Imo `-1` or `None` are better sentinels here, feel free to disagree with me and ignore this one

src/ldp/utils.py
```diff
@@ -110,3 +113,54 @@ def format_error_details(error: Exception) -> str:
...
+    - a ToolResponseMessage to ToolRequestMessage transition occurs
+
+    Args:
+        list_of_messages: The list of messages to break down.
+
+    Returns:
+        A list of transitions blocks, corresponding to trajectory transition.
+    """
+    if not list_of_messages:
+        return []
```
jamesbraza, line 130: Can we just design this function to be a no-op if `list_of_messages` is an empty list?

src/ldp/agent/simple_agent.py
sidnarayanan: Agreed, let's do None. 0 is a valid value - gives us a stateless agent.

src/ldp/agent/simple_agent.py
sidnarayanan: I prefer the classes - what's wrong with using types? We use typing to distinguish between tool request messages, tool response messages, env state messages, etc.

src/ldp/agent/simple_agent.py
```diff
@@ -19,6 +21,10 @@ class HiddenEnvStateMessage(EnvStateMessage):
     content: str = "[Previous environment state - hidden]"
 
 
+class HiddenTransitionsMessage(Message):
+    content: str = "[Previous transitions - hidden]"
```
sidnarayanan: "transitions" is an implementation detail the LLM is unaware of. I would simply say "Previous messages".

src/ldp/agent/simple_agent.py
```diff
@@ -125,6 +143,11 @@ class SimpleAgent(BaseModel, Agent[SimpleAgentState]):
         description="See SimpleAgentState.hide_old_action_content.",
     )
 
+    sliding_window: int = Field(
+        default=0,
+        description="Number of previous trajectory transitions to keep in the state. 0 means all previous transitions.",
```
sidnarayanan: Let's follow the description pattern used for the other fields.

src/ldp/utils.py
albertbou92: wait isn't it the same returning `list_of_messages` than `[]` in that case?

src/ldp/agent/simple_agent.py
albertbou92: ok lets go with None then

src/ldp/utils.py
```diff
@@ -110,3 +113,54 @@ def format_error_details(error: Exception) -> str:
...
+
+    filtered_messages: list[list[Message]] = []
+    i = 0
+
+    # First block: collect all system messages + user message
+    system_block: list[Message] = []
+    while i < len(list_of_messages) and list_of_messages[i].role in {"system", "user"}:
+        system_block.append(list_of_messages[i])
+        i += 1
```
sidnarayanan:
```suggestion
    for msg in list_of_messages:
        if msg.role not in {'system', 'user'}:
            break
        system_block.append(msg)
```

I found it harder to follow the index logic

src/ldp/utils.py
```diff
@@ -110,3 +113,54 @@ def format_error_details(error: Exception) -> str:
...
+    filtered_messages.append(system_block)
+
+    # Process all remaining messages in
+    # (ToolRequestMessage's, ToolResponseMessage's, EnvStateMessage (optional)) blocks
+    current_block: list[Message] = []
+    for j in range(i, len(list_of_messages)):
```
sidnarayanan:
```suggestion
    for j in range(len(system_block), len(list_of_messages)):
```

If you go with my suggestion to drop `i`

src/ldp/utils.py
```diff
@@ -110,3 +113,54 @@ def format_error_details(error: Exception) -> str:
...
+        msg = list_of_messages[j]
+        if (
+            j > 0
```
sidnarayanan: Worth a comment why `j=0` is a special case.

src/ldp/agent/react_agent.py
albertbou92: makes sense for description, but can we define default independently?

src/ldp/agent/simple_agent.py
albertbou92: is there an info specific `Message`?

src/ldp/utils.py
```diff
@@ -110,3 +113,54 @@ def format_error_details(error: Exception) -> str:
...
+
+    # First block: collect all system messages + user message
+    system_block: list[Message] = []
+    while i < len(list_of_messages) and list_of_messages[i].role in {"system", "user"}:
+        system_block.append(list_of_messages[i])
+        i += 1
+    filtered_messages.append(system_block)
+
+    # Process all remaining messages in
+    # (ToolRequestMessage's, ToolResponseMessage's, EnvStateMessage (optional)) blocks
```
sidnarayanan: This assumes a very specific structure that not all envs obey - some envs inject (non-EnvState) Messages in addition to the ToolResponseMessage.

It's more robust to assume each `ToolRequestMessage` is the start of a transition, since we enforce that here: https://github.com/Future-House/ldp/blob/dbc54f88f1af500efdb684ee11b831c19e12e89f/src/ldp/agent/agent.py#L51

All messages following a tool request (until the next one) can be assumed to be part of that transition.

src/ldp/utils.py
jamesbraza: Feel free to ignore this then

src/ldp/agent/simple_agent.py
albertbou92: as discussed in person, using  a `Message` here

src/ldp/utils.py
albertbou92: Actually is not needed

src/ldp/agent/react_agent.py
```diff
@@ -117,6 +117,16 @@ class ReActAgent(BaseModel, Agent[SimpleAgentState]):
         description="See SimpleAgentState.hide_old_env_states.",
     )
 
+    hide_old_action_content: bool = Field(
+        default=False,
+        description="See SimpleAgentState.hide_old_action_content.",
+    )
+
+    sliding_window: int | None = Field(
+        default=None,
```
jamesbraza, line 126:
```suggestion
        default=SimpleAgentState.model_fields["sliding_window"].default,
```

src/ldp/agent/simple_agent.py
```diff
@@ -71,6 +77,16 @@ def get_next_state(
             if hide_old_env_states is not None
             else self.hide_old_env_states
         )
+        if self.sliding_window is not None and self.sliding_window > 0:
+            msg_blocks = split_message_transitions(old_messages)
+            old_messages = (
+                msg_blocks[0]  # keep system messages + user message
```
jamesbraza, line 83: Won't this hit an `IndexError` if `old_messages` is empty? Also, feel free to ignore this one, if it won't happen

jamesbraza (approved):

src/ldp/agent/react_agent.py
jamesbraza: Feel free to ignore this one

src/ldp/agent/simple_agent.py
albertbou92: It should never be empty, `old_messages` will contain at least the initial user prompt/query.
The first block from split_message_transitions  is sys_prompt (if exists) + initial user prompt/query.

PR #337, opened 2025-09-16
Sliding window support fixes

albertbou92:
This PR fixes 2 issues from https://github.com/Future-House/ldp/pull/336

It also adds a unit test, which the previous PR probably should have included.

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -843,8 +843,16 @@ async def test_sliding_window():
 
     # Check that old messages are hidden
     messages = state.get_next_state(obs).messages
-    assert len(messages) == 4
-    assert messages[0].role == "user"
-    assert messages[1].content == "[Previous messages - hidden]"
-    assert isinstance(messages[2], ToolRequestMessage)
-    assert isinstance(messages[3], ToolResponseMessage)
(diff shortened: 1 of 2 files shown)


tests/test_agents.py
```diff
@@ -817,3 +824,27 @@ async def test_interactive(dummy_env: DummyEnv, mocker):
...
+        action, state, _ = await agent.get_asv(state, obs)
+        obs, _, _, _ = await env.step(action.value)
+
+    # Check that old messages are hidden
+    messages = state.get_next_state(obs).messages
+    assert len(messages) == 4
+    assert messages[0].role == "user"
+    assert messages[1].content == "[Previous messages - hidden]"
+    assert isinstance(messages[2], ToolRequestMessage)
+    assert isinstance(messages[3], ToolResponseMessage)
```
jamesbraza: Should we assert on the content in all of the messages? These assertions seem somewhat unspecific, maybe add failure messages explaining their rationale

```python
assert messages[1].content == "[Previous messages - hidden]", "Expected a hidden message because XYZ"
```

jamesbraza (approved):

PR #338, opened 2025-09-18
Reduced the react text and duplicated period

maykcaldas:
We were formatting the ReAct message to show the "Though: ... Action: ..." process in the message history. Because we added a period to this message, it resulted in a duplication in the final message (check image). This PR removes this duplication and simplified the message to avoid having this same template repeated all over

```
Based on this reasoning, let's select the appropriate tool!"
Action: "
```

image

### DIFF
--- a/src/ldp/graph/modules/react.py
+++ b/src/ldp/graph/modules/react.py
@@ -313,10 +313,12 @@ def postprocess_and_concat_resoning_msg(
     msgs: Iterable[Message], react_message: Message
 ) -> Messages:
     reasoning = (react_message.content or "").removeprefix("Thought: ")
+    if reasoning and not reasoning.rstrip().endswith("."):
+        reasoning = reasoning.rstrip() + "."
     return [
         *msgs,
         Message(
-            content=(f"Thought: {reasoning}.\nAction: "),
+            content=(f"Thought: {reasoning}\nAction: "),
             # Role is 'assistant' here (normally 'user') since we use the model's reasoning to ask for an action.
             role="assistant",
         ),
--- a/src/ldp/graph/modules/react.py
+++ b/src/ldp/graph/modules/react.py
@@ -312,16 +312,9 @@ class ReActModuleSinglePrompt:
 def postprocess_and_concat_resoning_msg(
     msgs: Iterable[Message], react_message: Message
 ) -> Messages:
-    reasoning = (react_message.content or "").removeprefix("Thought: ")
-    if reasoning and not reasoning.rstrip().endswith("."):
-        reasoning = reasoning.rstrip() + "."
     return [
         *msgs,
-        Message(
-            content=(f"Thought: {reasoning}\nAction: "),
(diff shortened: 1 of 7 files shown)


src/ldp/graph/modules/react.py
```diff
@@ -313,14 +313,12 @@ def postprocess_and_concat_resoning_msg(
...
+        reasoning = reasoning.rstrip() + "."
     return [
         *msgs,
         Message(
-            content=(
-                f"Thought: {reasoning}."
-                " Based on this reasoning, let's select the appropriate tool!"
-                "\nAction: "
-            ),
+            content=(f"Thought: {reasoning}\nAction: "),
```
sidnarayanan: Let's drop Thought/Action too

src/ldp/graph/modules/react.py
```diff
@@ -313,14 +313,12 @@ def postprocess_and_concat_resoning_msg(
     msgs: Iterable[Message], react_message: Message
 ) -> Messages:
     reasoning = (react_message.content or "").removeprefix("Thought: ")
+    if reasoning and not reasoning.rstrip().endswith("."):
```
jamesbraza: Can you add a comment why we need the period?

src/ldp/graph/modules/react.py
jamesbraza: Just so I understand, wouldn't that make it not ReAct tho?

sidnarayanan: Yeah we're straying from the paper, but IIRC the paper did its experiments on non-instruction-tuned models. This formatting doesn't really make sense in the modern day. It's on my to-do list to fully deprecate ReAct.

src/ldp/graph/modules/react.py
```diff
@@ -312,18 +312,9 @@ async def __call__(
 def postprocess_and_concat_resoning_msg(
     msgs: Iterable[Message], react_message: Message
 ) -> Messages:
-    reasoning = (react_message.content or "").removeprefix("Thought: ")
```
jamesbraza, line 315: @maykcaldas do you mind adding to the PR description the scenario we hit? I am trying to understand the duplicated period thing

src/ldp/graph/modules/react.py
maykcaldas: We still add the explanation that the LLM should reason and take action in the prompt. This is just not injecting this message into the trajectory. I think this is still ReAct because we still have the "though > action" process. It's just being explicitly saved in the message history

src/ldp/graph/modules/react.py
maykcaldas: I removed this check. I just wanted to make sure the period was there. The previous message had it like f"{reasoning}." and was creating a message with two periods because reasoning already had one. Now we just add the `react_message` directly to the message history

src/ldp/graph/modules/react.py
maykcaldas: Edited it

PR #340, opened 2025-10-01
EDI-45 fix: cost callbacks fire only on final chunk with usage (EDI-51)[https://linear.app/edisonscientific/issue/EDI-45/logs-are-flooded-with-duplicate-cost-tracking-events]

jabra:


packages/lmi/src/lmi/cost_tracker.py
```diff
@@ -139,7 +139,8 @@ def __aiter__(self):
 
     async def __anext__(self):
         response = await self.stream.__anext__()
-        if GLOBAL_COST_TRACKER.enabled.get():
+        # Only record costs on the final chunk (which contains usage information)
+        if GLOBAL_COST_TRACKER.enabled.get() and getattr(response, "usage", None):
```
jamesbraza: Can we move this logic into `record`? Imo it's not the job of `__anext__` to filter, let's encapsulate that

packages/lmi/tests/test_cost_tracking.py
```diff
@@ -245,3 +245,49 @@ async def async_callback(response):  # noqa: RUF029
...
+    @pytest.mark.asyncio
+    async def test_stream_wrapper_only_records_final_chunk(self):
+        """Test that cost callbacks are only fired on the final chunk with usage info."""
+        # Create mock stream that yields 3 chunks: 2 intermediate, 1 final
+        intermediate_chunk_1 = MagicMock(usage=None)
+        intermediate_chunk_2 = MagicMock(usage=None)
+        final_chunk = MagicMock(usage=MagicMock(prompt_tokens=10, completion_tokens=20))
+
+        mock_stream = MagicMock()
+        mock_stream.__anext__ = AsyncMock(
```
jamesbraza:
```
        mock_stream = MagicMock(__anext__= AsyncMock(
```

We can do this in one line, fwiw

packages/lmi/tests/test_cost_tracking.py
```diff
@@ -245,3 +245,49 @@ async def async_callback(response):  # noqa: RUF029
...
+        with (
+            cost_tracking_ctx(),
+            patch("litellm.cost_calculator.completion_cost", return_value=0.01),
+        ):
+            # Consume all chunks
+            chunks = []
+            try:
+                while True:
+                    chunk = await anext(wrapper)
+                    chunks.append(chunk)
```
jamesbraza: Can we not do an `async for` here?

packages/lmi/src/lmi/cost_tracker.py
jabra: makes sense, updated

packages/lmi/tests/test_cost_tracking.py
```diff
@@ -245,3 +245,51 @@ async def async_callback(response):  # noqa: RUF029
...
+        GLOBAL_COST_TRACKER.add_callback(async_callback)
+
+        with (
+            cost_tracking_ctx(),
+            patch("litellm.cost_calculator.completion_cost", return_value=0.01),
+        ):
+            # Consume all chunks
+            chunks = []
+            async for chunk in wrapper:
+                chunks.append(chunk)
```
jamesbraza: I think you can do this:

```suggestion
            chunks = [chunk async for chunk in wrapper]
```

packages/lmi/src/lmi/cost_tracker.py
```diff
@@ -26,6 +26,10 @@ def add_callback(self, callback: Callable[[LLMResponse], Awaitable]) -> None:
         self._callbacks.append(callback)
 
     async def record(self, response: LLMResponse) -> None:
+        # Only record costs on responses with usage information
+        if not getattr(response, "usage", None):
+            return
```
jamesbraza: Now seeing this, I realize `litellm.cost_calculator.completion_cost` is below this. It feels like we should not be manually inspecting `response.usage` when we've outsourced cost computation to `litellm`.

Is there any way we can do this upon the return of `litellm.cost_calculator.completion_cost`?

packages/lmi/tests/test_cost_tracking.py
```diff
@@ -245,3 +245,51 @@ async def async_callback(response):  # noqa: RUF029
             assert len(callback_calls) == 1
             assert callback_calls[0] == mock_response
             assert GLOBAL_COST_TRACKER.lifetime_cost_usd > 0
+
+    @pytest.mark.asyncio
+    async def test_stream_wrapper_only_records_final_chunk(self):
```
jamesbraza, line 248: Okay so this test is fine, but really it's just geared towards completions. I am not sure we've covered all cases yet with the change here.

Can we move this to a parameterized test with some OpenAI and some Anthropic model that uses `pytest.mark.vcr` to cover:
- Completion model: streaming and non-streaming
- Embedding model: streaming and non-streaming

packages/lmi/src/lmi/cost_tracker.py
jabra: We check for usage presence rather than cost > 0 because:
 - Free models, unknown models, or custom pricing can have cost = 0
 - We still want to fire callbacks for these to maintain visibility

jamesbraza: Alright sounds good, thanks for looking into it!

packages/lmi/src/lmi/cost_tracker.py
```diff
@@ -26,6 +26,13 @@ def add_callback(self, callback: Callable[[LLMResponse], Awaitable]) -> None:
         self._callbacks.append(callback)
 
     async def record(self, response: LLMResponse) -> None:
+        # Only record on responses with usage information (final chunk in streaming)
+        # We check for usage presence rather than cost > 0 because:
+        # - Free models, unknown models, or custom pricing can have cost = 0
+        # - We still want to fire callbacks for these to maintain visibility
+        if not getattr(response, "usage", None):
```
jamesbraza, line 33:
```suggestion
        if not hasattr(response, "usage"):
```

This may be better because it avoids a truthy check on `.usage` when present. Also, feel free to ignore this one

jamesbraza (approved):
Nice work, LGTM!

packages/lmi/src/lmi/cost_tracker.py
jabra: it is because of this option

stream_options = {
            "include_usage": True,
} 
...
stream_completions = await track_costs_iter(self.router.acompletion)(
            self.name,
            prompts,
            stream=True,
            stream_options=stream_options,
            **kwargs,
        )

when it is true you get usage: null for every intermediate chunk

jamesbraza: Ah I see, then yeah we nees the truthy check. Thanks for answering

PR #446, opened 2026-04-18
Moving things into LMI

sidnarayanan:
Bringing some functionality up from LiteLLM into LMI:
- Model configuration uses opinionated pydantic models (`ModelSpec, LLMConfig`), which get translated into litellm config dicts just in time
- LMI handles retries and fallbacks. We use slightly more intelligent retries - only on things that are actually worth retrying, and with jittered exp backoff
- Because we own retries/fallbacks, we have better logging
- `ModelSpec.responses_api` lets us toggle-on Responses backend selectively. This means we can swap between the backends in a fallback cascade
- Response validation is pushed down from LDP into LMI, so it can use the new LMI retry/fallback logic.

Notably, this drops all dependence on `litellm.Router` except for embeddings, which we can likely handle in a similar way. It also gets rid of a few things that annoyed me:
- Having to remember whether parameters go in `model_list` or top-level kwargs
- Differently-shaped configuration at `SimpleAgent/LLMCallOp/LiteLLMModel/litellm` levels. Now, the first 3 all operate on `LLMConfig`.

Some proof that this is backwards compatible: no cassette churn besides from new/dropped tests.


packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,279 @@
...
+
+ResponseValidator = Callable[[LLMResult], Awaitable[None] | None]
+
+_DEFAULT_TEMPERATURE = 1.0
+_OPENAI_ONLY_PARAMS = frozenset({"logprobs", "top_logprobs"})
+
+# Per-call retry kwargs that LiteLLM honors via its own internal retry loop. LMI
+# owns retries through `_run_with_fallbacks` + `ModelSpec.max_retries`, so these
+# must never reach `litellm.acompletion`/`litellm.aresponses` regardless of how
+# they ended up in `ModelSpec.extra_params`.
```
jamesbraza:
```suggestion
# they ended up in our `ModelSpec.extra_params`.
```

Felt we should clarify that `ModelSpec` is our thing

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+class ModelSpec(BaseModel):
+    """One model in an `LLMConfig` chain."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    name: str = Field(
+        description=(
+            "LiteLLM model string, e.g. 'gpt-4o-mini' or 'claude-3-5-sonnet-20241022'."
```
jamesbraza: But don't LiteLLM model string start with like `openai/gpt-4o-mini`? Maybe clarify this.

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,279 @@
...
+                    f"{key!r} is only supported on OpenAI models; got {name!r}."
+                )
+            else:
+                extra_overrides[key] = value
+
+        spec_field_overrides.setdefault("name", name)
+        merged_extra = (
+            extra | extra_overrides | spec_field_overrides.pop("extra_params", {})
+        )
+        return cls(extra_params=merged_extra, **spec_field_overrides)
```
jamesbraza, line 145: Should we do this validation in a `field`/`model_validator`, so it applies to all `ModelSpec`, not just this `classmethod`?

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,279 @@
...
+
+        primary_name = model_list[0]["model_name"]
+        ordered: list[str] = [primary_name, *fallback_map.get(primary_name, [])]
+        for name in params_by_name:
+            if name not in ordered:
+                ordered.append(name)
+
+        router_kwargs = legacy.get("router_kwargs") or {}
+        default_timeout = router_kwargs.get("timeout", 60.0)
+        default_retries = router_kwargs.get("num_retries", 3)
```
jamesbraza: Any chance we can not inject `, 60.0` and `, 3` defaults here? The less injection we have, the better

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,279 @@
...
+        if "gemini" in name:
+            extra["safety_settings"] = DEFAULT_VERTEX_SAFETY_SETTINGS
+        extra["temperature"] = _DEFAULT_TEMPERATURE
+
+        spec_field_overrides: dict[str, Any] = {}
+        extra_overrides: dict[str, Any] = {}
+        for key, value in overrides.items():
+            if key in cls.model_fields:
+                spec_field_overrides[key] = value
+            elif key in _OPENAI_ONLY_PARAMS and not is_openai:
```
jamesbraza: Gemini supports these too, can you have Gemini work with these too?

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,279 @@
...
+        Applies: Gemini default safety settings; `temperature` / `max_tokens`
+        defaults; and silent drop of `logprobs` / `top_logprobs` for non-OpenAI
+        providers (which don't support them). Explicit values in `overrides`
+        always win over the defaults.
+
+        `overrides` may set any `ModelSpec` field, plus request-shape kwargs
+        (`temperature`, `max_tokens`, `n`, `logprobs`, `top_logprobs`,
+        `safety_settings`, ...) which are merged into `extra_params`.
+        """
+        is_openai = _is_openai_provider(name)
```
jamesbraza: Consider using `litellm` for this instead of a DIY'd helper, it can detect this using `litellm.get_llm_provider`

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,279 @@
...
+_RESERVED_LEGACY_PARAMS = frozenset({
+    "model",
+    "api_base",
+    "api_key",
+    "timeout",
+    "max_retries",
+})
+
+
+def _spec_from_legacy_params(
```
jamesbraza: Can we make this a `classmethod` of `ModelSpec`? Then we can get `Self` as the return type

packages/lmi/src/lmi/constants.py
```diff
@@ -1,8 +1,5 @@
-import os
 from sys import version_info
 
-USE_RESPONSES_API = os.environ.get("USE_RESPONSES_API", "").lower() in {"1", "true"}
```
jamesbraza, line 4: Nice this is awesome

packages/lmi/pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
     "coredis>=3.0.1",  # Lower pin v3 for Redis class, pin v3.0.1 for fix in https://github.com/alisaifee/coredis/commit/7e9b1a1b384cd97725cab479d2ce091e3b0823d2
     "fhaviary>=0.14.0",  # For multi-image support
     "limits[async-redis]>=4.8",  # Specify 'async-redis' since that's what we use. Lower pin for RedisBridge.key_prefix.
-    "litellm>=1.81.10,<=1.82.6",  # Lower pin for MAX_CALLBACKS refactor from https://github.com/BerriAI/litellm/pull/20781, upper pin for supply chain attack
+    "litellm>=1.83.10",  # Lower pin for MAX_CALLBACKS refactor from https://github.com/BerriAI/litellm/pull/20781, upper pin for supply chain attack
```
jamesbraza: I think rebase atop `main` as this has been changed a bit

agolajko: done

packages/lmi/src/lmi/config.py
agolajko: solved

packages/lmi/src/lmi/config.py
agolajko: _unsupported_params now uses litellm.get_supported_openai_params rather than a DIY

packages/lmi/src/lmi/config.py
agolajko: removed

packages/lmi/src/lmi/config.py
agolajko: solved

packages/lmi/src/lmi/config.py
agolajko: solved

packages/lmi/src/lmi/config.py
agolajko: added

packages/lmi/src/lmi/config.py
agolajko: yes

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,343 @@
...
+
+class ModelSpec(BaseModel):
+    """One model in an `LLMConfig` chain."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    name: str = Field(
+        description=(
+            "LiteLLM model string in 'provider/model' format, "
+            "e.g. 'openai/gpt-4o-mini' or 'anthropic/claude-3-5-sonnet-20241022'."
```
jamesbraza: [nit] Can you use `example=["openai/gpt-4o-mini"]`

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,343 @@
...
+        ),
+    )
+    api_base: str | None = None
+    api_key: SecretStr | None = None
+    timeout: float = Field(default=60.0, description="Per-request timeout in seconds.")
```
jamesbraza: Can we allow `float | None`, I don't think we need to always require a timeout

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,343 @@
...
+    max_retries: int = Field(
+        default=3,
+        description=(
+            "Retries against this model before falling over to the next entry"
+            " in the chain."
```
jamesbraza: Can you document what `max_retries=0` would do?

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,343 @@
...
+        exclude=True,
+        description=(
+            "Custom parser for converting LLM completions to tool requests."
+            " Lifted onto `LiteLLMModel.tool_parser` during model construction;"
+            " see that field for the accepted signatures."
+        ),
+    )
+
+    @classmethod
+    def coerce(cls, v: Any) -> LLMConfig:
```
jamesbraza: The idiomatic Pydantic way to do this is a `model_validator(mode="before")`, can we move to that? Then we can ditch `LLMConfigField` below I think

packages/lmi/src/lmi/retry.py
```diff
@@ -0,0 +1,65 @@
...
+using `BACKOFF_INITIAL` / `BACKOFF_CAP` exponential backoff with full jitter.
+When retries are exhausted or the error indicates that another model might
+succeed, it advances to the next entry in the chain.
+
+Classification is delegated to litellm's exception hierarchy: `_RETRYABLE`
+covers transient failures; `_FALLBACKABLE` covers errors that are stable
+against retry but may resolve on a different model.
+"""
+
+from __future__ import annotations
```
jamesbraza: Do we need this? Can we drop it? 🤓

packages/lmi/src/lmi/litellm_patches.py
```diff
@@ -39,41 +36,7 @@ def _patched_model_dump(self, *args, by_alias=None, **kwargs):
...
 
-# Patch 2: Provider-specific 400 error retry
-#
-# Issue: Anthropic has a 100-image limit that returns 400 Bad Request. litellm's
-# default behavior doesn't retry 400 errors, so requests with >100 images fail
-# without trying Gemini/OpenAI which could handle them.
-#
-# This patch checks for specific provider-limit error messages and allows retry
-# only for those, not for genuine client errors (malformed requests, etc.).
-def _apply_should_retry_patch():
```
jamesbraza, line 50: Why are we dropping this patch? It seems like we can still hit this 100-image limit?

packages/lmi/src/lmi/retry.py
```diff
@@ -0,0 +1,65 @@
...
+from lmi.exceptions import ModelRefusalError, ResponseValidationError
+
+BACKOFF_INITIAL = 1.0
+BACKOFF_CAP = 30.0
+
+
+_RETRYABLE: tuple[type[BaseException], ...] = (
+    litellm.RateLimitError,
+    litellm.Timeout,
+    litellm.APIConnectionError,
```
jamesbraza, line 36: What about `litellm`'s `BadGatewayError`?

packages/lmi/src/lmi/retry.py
```diff
@@ -0,0 +1,65 @@
...
+    litellm.InternalServerError,
+    litellm.ServiceUnavailableError,
+    ResponseValidationError,
+)
+
+_FALLBACKABLE: tuple[type[BaseException], ...] = (
+    litellm.ContextWindowExceededError,
```
jamesbraza: Note that this is a subclass of `litellm.BadRequestError` below. Maybe we:
1. Remove `litellm.BadRequestError` subclasses
2. Document them with `litellm.BadRequestError` below

packages/lmi/src/lmi/retry.py
```diff
@@ -0,0 +1,65 @@
...
+
+def should_fallback(exc: BaseException) -> bool:
+    """True if another model might succeed where this one failed."""
+    return isinstance(exc, _FALLBACKABLE)
+
+
+def backoff_seconds(attempt: int) -> float:
+    """Exponential backoff with full jitter. `attempt` is 0-indexed."""
+    ceiling = min(BACKOFF_CAP, BACKOFF_INITIAL * (2**attempt))
+    return random.uniform(0, ceiling)
```
jamesbraza: Why are we not using `tenacity` for this? This is built-in there

packages/lmi/src/lmi/retry.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def should_retry(exc: BaseException) -> bool:
+    """True if the same model might succeed on another try."""
+    return isinstance(exc, _RETRYABLE)
+
+
+def should_fallback(exc: BaseException) -> bool:
+    """True if another model might succeed where this one failed."""
+    return isinstance(exc, _FALLBACKABLE)
```
jamesbraza: Can we not use `tenacity` for this? I am pretty sure this is built into `tenacity`

packages/lmi/src/lmi/retry.py
```diff
@@ -0,0 +1,65 @@
...
+    litellm.APIConnectionError,
+    litellm.InternalServerError,
+    litellm.ServiceUnavailableError,
+    ResponseValidationError,
+)
+
+_FALLBACKABLE: tuple[type[BaseException], ...] = (
+    litellm.ContextWindowExceededError,
+    litellm.NotFoundError,
+    litellm.ContentPolicyViolationError,
```
jamesbraza: Same here, isn't this `litellm.BadRequestError` so we don't need this one, instead just document below

packages/lmi/src/lmi/llms.py
```diff
@@ -1110,60 +1071,112 @@ async def check_rate_limit(self, token_count: float, **kwargs) -> None:
...
+        llm_config = cast("LLMConfig", self.llm_config)  # populated by after-validator
+        last_exc: BaseException | None = None
+        for spec_idx, spec in enumerate(llm_config.models):
+            for i in range(spec.max_retries + 1):
+                try:
+                    result = await attempt(spec, *args, **kwargs)
+                except Exception as exc:
+                    last_exc = exc
+                    if should_retry(exc) and i < spec.max_retries:
+                        delay = backoff_seconds(i)
```
jamesbraza: Why not just use `tenacity` for this:

```python
for attempt in Retrying(...):
    with attempt:
        ...
```

packages/lmi/src/lmi/config.py
agolajko: added

packages/lmi/src/lmi/config.py
agolajko: added

packages/lmi/src/lmi/config.py
agolajko: done, also added pydantic validation so its ge=0

packages/lmi/src/lmi/config.py
agolajko: Thanks for the suggestion, done

Edited other files that used the LLMConfigField

packages/lmi/src/lmi/retry.py
agolajko: done

packages/lmi/src/lmi/litellm_patches.py
agolajko: 100-image-limit case is now handled in [packages/lmi/src/lmi/retry.py](https://github.com/Future-House/ldp/pull/446/changes#diff-a6b222b56a60af2f9a8147ddd253a6fecf69f3d8248e907b284cab87bd739674R46): _PROVIDER_LIMIT_PATTERNS = ("too much media",) plus should_fallback() matches that on a plain BadRequestError

packages/lmi/src/lmi/retry.py
agolajko: added

packages/lmi/src/lmi/retry.py
agolajko: The inner loop is now tenacity-driven [retry.py:85](https://github.com/Future-House/ldp/pull/446/changes#diff-a6b222b56a60af2f9a8147ddd253a6fecf69f3d8248e907b284cab87bd739674R85) and retries the same model; the outer loop is our own iteration over LLMConfig.models llms.py:1131 _run_with_fallbacks

packages/lmi/src/lmi/llms.py
agolajko: done

packages/lmi/src/lmi/retry.py
agolajko: I re-narrowed the policy and brought the subclasses back to being meaningful

packages/lmi/src/lmi/retry.py
agolajko: Tenacity drives the inner retry loop

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,348 @@
...
+
+# Per-call retry kwargs that LiteLLM honors via its own internal retry loop. LMI
+# owns retries through `_run_with_fallbacks` + `ModelSpec.max_retries`, so these
+# must never reach `litellm.acompletion`/`litellm.aresponses` regardless of how
+# they ended up in our `ModelSpec.extra_params`.
+_LITELLM_RETRY_KWARGS = frozenset({"num_retries", "max_retries"})
+
+# `tool_parser` is an LMI-level callable handled on `LiteLLMModel`; if it ever
+# ends up in `extra_params` (e.g. via a name-shape dict that wasn't unpacked
+# by the `LLMConfig` before-validator), filter it before dispatching to litellm.
```
jamesbraza: > e.g. via a name-shape dict that wasn't unpacked by the `LLMConfig` before-validator

This is slop-ish, consider cleaning this up

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,348 @@
...
+    """One model in an `LLMConfig` chain."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    name: str = Field(
+        description=(
+            "LiteLLM model string in 'provider/model' format, "
+            "e.g. 'openai/gpt-4o-mini' or 'anthropic/claude-3-5-sonnet-20241022'."
+        ),
+        examples=["openai/gpt-4o-mini"],
```
jamesbraza: If you put `examples` (and thanks for that), remove the "e.g." in the `description`

packages/lmi/src/lmi/config.py
```diff
@@ -0,0 +1,348 @@
...
+    @classmethod
+    def from_legacy_params(
+        cls, params: dict[str, Any], router_kwargs: dict[str, Any]
+    ) -> Self:
+        """Build a spec from a legacy `litellm_params` entry plus router defaults.
+
+        `params` is one `model_list[*].litellm_params` dict; `router_kwargs` is
+        the config's `router_kwargs`. Only forwards `timeout`/`max_retries` when
+        present so this class's own field defaults remain the single source of
+        truth (no re-injected literals).
```
jamesbraza: > (no re-injected literals).

What does this mean? De-slop this

packages/lmi/src/lmi/litellm_patches.py
jamesbraza: Sounds good, thanks

packages/lmi/src/lmi/llms.py
```diff
@@ -871,6 +873,18 @@ async def aresponses(self, *args, **kwargs):
         return await litellm.aresponses(*args, **(self._default_kwargs | kwargs))
 
 
+def _strip_tool_only_kwargs(call_kwargs: dict[str, Any]) -> None:
```
jamesbraza: Why do we need this, can we drop this? I don't like secretly manipulating stuff like this, I would rather callers just call things correctly.

Stuff like this, especially manipulating caller's mutable inputs, can cause big headaches

packages/lmi/src/lmi/llms.py
```diff
@@ -1110,60 +1083,105 @@ async def check_rate_limit(self, token_count: float, **kwargs) -> None:
...
+        """
+        llm_config = cast("LLMConfig", self.llm_config)  # populated by after-validator
+        last_exc: BaseException | None = None
+        for spec_idx, spec in enumerate(llm_config.models):
+            try:
+                async for retry_attempt in model_retrying(spec.max_retries):
+                    # `return` lives inside `with` so it runs only on success;
+                    # on failure tenacity's context manager swallows the error
+                    # (to retry) and execution would otherwise fall through with
+                    # `result` unbound.
```
jamesbraza: [nit] this could use some de-slopping. I would rather see localized one-line comments, maybe split this up

packages/lmi/src/lmi/llms.py
```diff
@@ -1380,34 +1419,52 @@ async def acompletion_iter(
...
             call_kwargs["previous_response_id"] = previous_response_id
 
-        response = await track_costs(router.aresponses)(**call_kwargs)
+        try:
+            response = await track_costs(litellm.aresponses)(**call_kwargs)
+        except Exception:
+            logger.exception("aresponses attempt failed on %s.", spec.name)
+            raise
 
-        used_model = response.model or self.name
```
jamesbraza, line 1410: Why are you removing `or self.name`? Just making sure that's not a regression

packages/lmi/src/lmi/retry.py
```diff
@@ -0,0 +1,99 @@
...
+    litellm.InternalServerError,
+    litellm.ServiceUnavailableError,
+    litellm.BadGatewayError,
+    ResponseValidationError,
+)
+
+# 400s that are genuinely provider-specific (a sibling model may accept the same
+# input) rather than a malformed request. litellm surfaces these as a plain
+# `BadRequestError`, so we match on the message; see `should_fallback`.
+_PROVIDER_LIMIT_PATTERNS = ("too much media",)
```
jamesbraza, line 47: Can you preserve info that this comes from Anthropic, in a comment somehow? Provider-knowledge is useful

jamesbraza (approved):
Huge refactor, wow. Nice work! Looking forward to using `LLMConfig`

packages/lmi/src/lmi/retry.py
agolajko: added

packages/lmi/src/lmi/llms.py
agolajko: thanks for the spot

packages/lmi/src/lmi/llms.py
agolajko: yep, need this, but I made it more explicit now

PR #452, opened 2026-04-23
lmi: support TLS Redis (rediss://) in GlobalRateLimiter

jabra:
## Summary

`GlobalRateLimiter` previously hardcoded `async+redis://` for the storage client (`rate_limiter.py:155`) and opened the `get_rate_limit_keys` SCAN client without any `ssl=` kwarg. Against a TLS-only Redis — e.g. AWS ElastiCache with in-transit encryption enabled — the TCP handshake succeeds but the server then hangs waiting for a ClientHello that never arrives. Symptom in the wild: `get_rate_limit_keys()` hangs during agent startup, blocking `_choose_model_based_on_limits` and every downstream rate-limited call.

This PR makes both Redis code paths scheme-aware and consistent:

- Two shared helpers — `_wants_tls(redis_url)` and `_strip_scheme(redis_url)` — decide TLS from an explicit `rediss://` / `redis://` scheme, or (for bare `host:port` URLs) from a `REDIS_USE_TLS` env var parsed as a whitelist (`1|true|yes|on`) rather than `bool()` (which would treat `"0"`/`"false"` as true).
- The storage property now builds `async+rediss://{host}:{port}` when TLS is required, instead of unconditionally prefixing `async+redis://`.
- `get_rate_limit_keys()` uses the same helpers and passes `ssl=<tls>` to the direct `coredis.Redis(...)` client.

## Behavior matrix

| `REDIS_URL`              | `REDIS_USE_TLS` | Storage URL                   | Scan client `ssl` |
|--------------------------|-----------------|-------------------------------|-------------------|
| `host:6379`              | unset           | `async+redis://host:6379`     | `False`           |
| `host:6379`              | `true`          | `async+rediss://host:6379`    | `True`            |
| `host:6379`              | `0` / `false`   | `async+redis://host:6379`     | `False`           |
| `redis://host:6379`      | any             | `async+redis://host:6379`     | `False`           |
| `rediss://host:6379`     | any             | `async+rediss://host:6379`    | `True`            |

The bare-host + no-env-var row is the GCP Memorystore path — unchanged.

## Test plan

- [x] `uv run --directory packages/lmi pytest tests/test_rate_limiter.py` → 38 passed
- [x] `uv run mypy packages/lmi/src/lmi/rate_limiter.py` → clean
- [x] `pre-commit run --files packages/lmi/src/lmi/rate_limiter.py packages/lmi/tests/test_rate_limiter.py` → all hooks pass (ruff, mypy, codespell, detect-secrets, blacken-docs)
- [x] New parametrized coverage for the scan client AND the storage URL across plain, `redis://`, `rediss://`, and both `REDIS_USE_TLS` directions (including the `"0"` regression case that `bool()` would have mishandled)
- [x] Validated end-to-end in a downstream service: a TLS-only ElastiCache endpoint that previously hung during `get_rate_limit_keys` now completes startup and proceeds to model selection.

## Follow-up

A downstream service (paperqa-server) currently carries a local override of `get_rate_limit_keys` that added `ssl=True` to the scan client. Once this lands and is released, that override can be dropped.

### DIFF
--- a/packages/lmi/src/lmi/rate_limiter.py
+++ b/packages/lmi/src/lmi/rate_limiter.py
@@ -107,12 +107,15 @@ class GlobalRateLimiter:
         Args:
             rate_config: Optional dictionary mapping (namespace, primary_key) tuples to
                 RateLimitItem instances. If not provided, the default RATE_CONFIG is used.
-            use_in_memory: If True, use in-memory storage instead of Redis, even if
-                Redis URL is available.
+            use_in_memory: If True, force in-memory storage even if a Redis URL
+                is available. When False (the default), in-memory storage is still
+                used as a fallback if no Redis URL is provided via `redis_url` or
+                the `REDIS_URL` environment variable.
             redis_url: Optional Redis URL to use for storage. If not provided, the
-                REDIS_URL environment variable will be used. This parameter allows
-                direct specification of the Redis URL without relying on environment
-                variables.
+                `REDIS_URL` environment variable will be used. Supports
+                `redis://host:port` (plaintext), `rediss://host:port` (TLS, e.g.
+                AWS ElastiCache with in-transit encryption), and bare `host:port`
+                (treated as plaintext).
         """
         self._rate_config = RATE_CONFIG if rate_config is None else rate_config
         self._redis_url = redis_url or os.environ.get("REDIS_URL")
--- a/packages/lmi/src/lmi/rate_limiter.py
+++ b/packages/lmi/src/lmi/rate_limiter.py
@@ -116,7 +116,10 @@ class GlobalRateLimiter:
         """
         self.rate_config = RATE_CONFIG if rate_config is None else rate_config
         self.use_in_memory = use_in_memory
-        self.redis_url = redis_url
+        self.redis_url = redis_url or os.environ.get("REDIS_URL", "")
+        self.redis_tls = self.redis_url.startswith("rediss://")
+        self.redis_scheme = "async+rediss" if self.redis_tls else "async+redis"
+        self.redis_bare_url = self._strip_scheme(self.redis_url)
         self._storage: RedisStorage | MemoryStorage | None = None
         self._rate_limiter: MovingWindowRateLimiter | None = None
         self._current_ip: str | None = None
@@ -134,25 +137,8 @@ class GlobalRateLimiter:
         return None
 
     @staticmethod
-    def _wants_tls(redis_url: str) -> bool:
-        if redis_url.startswith("rediss://"):
-            return True
-        if redis_url.startswith("redis://"):
-            return False
-        return os.environ.get("REDIS_USE_TLS", "").lower() in {
-            "1",
-            "true",
-            "yes",
-            "on",
-        }
-
-    @staticmethod
     def _strip_scheme(redis_url: str) -> str:
-        if redis_url.startswith("rediss://"):
-            return redis_url[len("rediss://") :]
-        if redis_url.startswith("redis://"):
-            return redis_url[len("redis://") :]
-        return redis_url
+        return redis_url.removeprefix("rediss://").removeprefix("redis://")
 
     async def outbount_ip(self) -> str:
         if self._current_ip is None:
@@ -171,13 +157,10 @@ class GlobalRateLimiter:
     @property
     def storage(self) -> RedisStorage | MemoryStorage:
         if self._storage is None:
-            redis_url = self.redis_url or os.environ.get("REDIS_URL")
-            if redis_url and not self.use_in_memory:
-                use_tls = self._wants_tls(redis_url)
-                bare_url = self._strip_scheme(redis_url)
-                scheme = "rediss" if use_tls else "redis"
-                self._storage = RedisStorage(f"async+{scheme}://{bare_url}")
-                logger.info("Connected to redis instance for rate limiting.")
+            if self.redis_url and not self.use_in_memory:
+                conn = f"{self.redis_scheme}://{self.redis_bare_url}"
+                self._storage = RedisStorage(conn)
+                logger.info(f"Connected to redis instance for rate limiting: {conn}")
             else:
                 self._storage = MemoryStorage()
                 logger.info("Using in-memory rate limiter.")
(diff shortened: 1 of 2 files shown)


packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -133,6 +133,27 @@ async def get_outbound_ip(session: aiohttp.ClientSession, url: str) -> str | Non
...
+        if redis_url.startswith("rediss://"):
+            return True
+        if redis_url.startswith("redis://"):
+            return False
+        return os.environ.get("REDIS_USE_TLS", "").lower() in {
+            "1",
+            "true",
+            "yes",
+            "on",
+        }
```
whitead: nit: I would expect the environ variable to override the implied tls setting from the URL

whitead (approved):

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -116,7 +116,10 @@ def __init__(
         """
         self.rate_config = RATE_CONFIG if rate_config is None else rate_config
         self.use_in_memory = use_in_memory
-        self.redis_url = redis_url
+        self.redis_url = redis_url or os.environ.get("REDIS_URL", "")
```
jamesbraza: Can you not inject `, ""`? I don't want us to have to think about empty string

Below in the original diff we had `os.environ.get("REDIS_URL", ":")`

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -116,7 +116,10 @@ def __init__(
...
+        self.redis_tls = self.redis_url.startswith("rediss://")
+        self.redis_scheme = "async+rediss" if self.redis_tls else "async+redis"
```
jamesbraza: Can you put a comment why `rediss` (two s)

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -133,6 +136,10 @@ async def get_outbound_ip(session: aiohttp.ClientSession, url: str) -> str | Non
             logger.warning(f"Error occurred while connecting to {url}.", exc_info=True)
         return None
 
+    @staticmethod
+    def _strip_scheme(redis_url: str) -> str:
+        return redis_url.removeprefix("rediss://").removeprefix("redis://")
```
jamesbraza, line 149: If this is only used in one place, can we just inline this code

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -269,42 +276,28 @@ async def get_rate_limit_keys(
...
-        # parse redis_url which may be "host:port" or ":password@host:port"
-        # the prefixed : looks like a bug waiting to happen but this is how
-        # the redis client handles uris without a username this way
-        try:
-            parsed = urlparse(f"redis://{redis_url}")
-        except ValueError as exc:
-            raise ValueError(
-                f"Failed to parse host and port from Redis URL {redis_url!r},"
-                " correctly pass at initialization or set env variable REDIS_URL."
-            ) from exc
```
jamesbraza, line 285: Why are you removing these?

packages/lmi/tests/test_rate_limiter.py
```diff
@@ -409,16 +409,27 @@ class TestGlobalRateLimiter:
...
+                id="rediss-scheme",
+            ),
         ],
     )
     @pytest.mark.asyncio
     async def test_get_rate_limit_keys_parses_redis_url(
         self, redis_url_factory, expected_redis_kwargs
     ) -> None:
-        """Test that get_rate_limit_keys correctly parses both plain and password-protected Redis URLs."""
+        """Test that get_rate_limit_keys parses Redis URLs and configures TLS correctly."""
```
jamesbraza: Why are you removing password mentions from the docstring?

jamesbraza (commented):
Nice work, though I think you clobbered a few things worth keeping around

packages/lmi/src/lmi/rate_limiter.py
ypicard: Keeps things a bit cleaner in the init. I'll keep it this way.

packages/lmi/src/lmi/rate_limiter.py
ypicard: I'll bring one back. They clutter the code with no real added value imo, but I may be wrong.

packages/lmi/tests/test_rate_limiter.py
ypicard: Reverted

packages/lmi/src/lmi/rate_limiter.py
jamesbraza: Oh if you think they're not adding value, then ditch em

packages/lmi/src/lmi/rate_limiter.py
ypicard: I updated this logic.

packages/lmi/tests/test_rate_limiter.py
```diff
@@ -409,9 +409,20 @@ class TestGlobalRateLimiter:
...
             ),
+            pytest.param(
+                lambda: "redis://10.58.188.212:6379",
+                {"host": "10.58.188.212", "port": 6379, "password": None, "ssl": False},
+                id="redis-scheme",
+            ),
+            pytest.param(
+                lambda: "rediss://10.58.188.212:6379",
+                {"host": "10.58.188.212", "port": 6379, "password": None, "ssl": True},
+                id="rediss-scheme",
```
jamesbraza, line 424: Maybe say `ssl-scheme`

packages/lmi/tests/test_rate_limiter.py
```diff
@@ -430,6 +441,34 @@ async def test_get_rate_limit_keys_parses_redis_url(
...
+            ),
+        ],
+    )
+    def test_storage_uses_expected_redis_scheme(
+        self, redis_url, expected_storage_url
+    ) -> None:
+        with patch("lmi.rate_limiter.RedisStorage") as mock_redis_storage:
+            limiter = GlobalRateLimiter(redis_url=redis_url)
+            _ = limiter.storage
```
jamesbraza, line 469: Why are we doing this? Maybe put an inline comment

jamesbraza (approved):
Nice work!

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -114,9 +114,17 @@ def __init__(
                 direct specification of the Redis URL without relying on environment
                 variables.
         """
-        self.rate_config = RATE_CONFIG if rate_config is None else rate_config
-        self.use_in_memory = use_in_memory
-        self.redis_url = redis_url
+        self._rate_config = RATE_CONFIG if rate_config is None else rate_config
+        self._redis_url = redis_url or os.environ.get("REDIS_URL")
+        self._use_in_memory = use_in_memory or not self._redis_url
```
ypicard, line 122: @jamesbraza This defaults to in memory when redis isn't configured.

jamesbraza: Oh sounds great.

Can you document that in the `Args` section of the docstring? Or feel free to skip.

jamesbraza (approved):

PR #457, opened 2026-05-22
DEBUG: capture mypy 2.1.0 INTERNAL ERROR traceback

jamesbraza:
Scratch PR to capture the `--show-traceback` output for the mypy 2.1.0 INTERNAL ERROR seen on PR #450.

**Do not merge.** Closing once the traceback is captured.

Base is `main` so the diff against PR #450 is exactly this one line. CI's pre-commit hook will now print the real traceback instead of the bare INTERNAL ERROR notice.


jamesbraza:
Closing scratch PR. Traceback captured (sqlite3.OperationalError: database is locked at mypy/metastore.py:174). This matches upstream issue python/mypy#13916 (canonical) / #21525 (recently-closed duplicate). Fix applied on PR #450 as require_serial: true on the mypy hook.

PR #460, opened 2026-06-17
fix(lmi): bound Redis I/O and fail open in GlobalRateLimiter [CLD-320]

jabra:
## Problem

PaperQA agent trajectories intermittently freeze forever. `GlobalRateLimiter.try_acquire()` gates every LLM/embedding call on a Redis `test()`/`hit()` (via `limits` over `coredis`). The coredis client was created with no socket timeout (`stream_timeout=None`), so when an idle Redis (AWS ElastiCache) connection is silently dropped, the `await` blocks indefinitely. The `acquire_timeout` only bounds the `asyncio.sleep` polling loop, not the Redis I/O, so it never fires. Confirmed via thread dumps (event loop parked), `/proc/net/tcp` (idle ESTABLISHED :6379 sockets), and ElastiCache metrics (Redis healthy/idle).

## Fix (all in `packages/lmi/src/lmi/rate_limiter.py`)

- **Bound Redis I/O:** pass `stream_timeout`/`connect_timeout` (`RATE_LIMITER_REDIS_OP_TIMEOUT`, default 5s) to `RedisStorage` -> `coredis.Redis.from_url`. coredis's default retry policy retries twice on a fresh pooled connection, so a stalled op surfaces as an error instead of hanging.
- **Fail open:** on `RedisError`/`StorageError`/`OSError` (retries exhausted = Redis unreachable), `try_acquire` lets the request through instead of propagating. After `RATE_LIMITER_REDIS_MAX_FAILURES` (default 3) consecutive failures the instance permanently degrades to `MemoryStorage` (per-process limiting) with a loud `logger.warning`. `test()`/`hit()` returning `False` (limit full) still waits — unchanged.
- **Proxy egress:** `outbound_ip()` uses `aiohttp.ClientSession(trust_env=True)` so `HTTP(S)_PROXY` is honored.

Deliberate fail-open: the rate limiter is a guardrail, not a correctness requirement — freezing a multi-hour trajectory is worse than one un-coordinated call. Logged loudly.

## Tests

Added to `packages/lmi/tests/test_rate_limiter.py`: fail-open on timeout/connection error (no hang, no raise), degrade to MemoryStorage after N failures, counter reset on success, wait-loop unchanged on `hit()==False`, and `RedisStorage` constructed with the timeouts.


packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -30,6 +33,33 @@
...
 
+# Socket-level timeout (seconds) for individual Redis ops. Without this, coredis is
+# created with stream_timeout=None, so a silently-dropped idle ElastiCache connection
+# makes the await block forever (the acquire_timeout only bounds the asyncio.sleep
+# polling loop, not the Redis I/O). coredis.Redis.from_url defaults to retrying twice
+# on (ConnectionError, TimeoutError) on a fresh pooled connection, so a stalled op is
+# bounded by roughly timeout * (1 + retries) before it surfaces as an error.
+RATE_LIMITER_REDIS_OP_TIMEOUT = float(
+    os.environ.get("RATE_LIMITER_REDIS_OP_TIMEOUT", "5")
+)
```
SamWright822: Can we make these vars instead of magic strings? like global default vars

SamWright822: oh nvm i guess this PR is prob being closed?

PR #463, opened 2026-06-23
Avoided deptry fails on packages [CLD-320]

maykcaldas:
The CI creates a /docs directory inside ldp/packages that uv was trying to interpret as a package and failing.
Additionally, deptry in ldp fails when scanning packages/*. I excluded this url as lmi has it's own deptry step


maykcaldas:
Apparently this was fixed on another PR

PR #465, opened 2026-07-02
fix(lmi): require explicit scheme in Redis URL for GlobalRateLimiter [CLD-311]

ypicard:
## Why

Rate-limited agents that talk to a TLS-required Redis (for example AWS ElastiCache with in-transit encryption) were failing at startup. The rate limiter decided TLS purely from the URL scheme, but a scheme-less `host:port` was silently treated as plaintext. A plaintext client against a TLS-only endpoint hangs on the handshake, so the failure surfaced much later as a `HELLO timed out` error mid-task rather than at connect time. On plaintext-tolerant Redis (such as GCP Memorystore) the same scheme-less URL happened to work, which is why this stayed hidden.

## What changed

- `GlobalRateLimiter` now validates the Redis URL at construction and rejects a scheme-less URL with a clear error, instead of quietly defaulting to plaintext.
- The scheme (`redis://` vs `rediss://`) is the single source of truth for TLS. Storage goes through `limits` as `async+<url>` and the introspection client uses `coredis.Redis.from_url`, so both derive TLS from the scheme.
- Removed the now-redundant scheme-stripping and manual `ssl=` plumbing.

## Notes

- This is a behavior change for any caller passing a bare `REDIS_URL` (for example existing plaintext GCP setups). They now need an explicit `redis://` prefix. That is intentional: it fails fast at construction rather than silently mid-task.
- Tests cover scheme-less rejection, the schemed URL passing through to coredis (including the TLS case), and a scheme-less URL still being allowed when in-memory storage is forced.

### DIFF
--- a/packages/lmi/src/lmi/rate_limiter.py
+++ b/packages/lmi/src/lmi/rate_limiter.py
@@ -30,6 +30,8 @@ CROSSREF_BASE_URL = f"https://{CROSSREF_HOST}"
 
 GLOBAL_RATE_LIMITER_TIMEOUT = float(os.environ.get("RATE_LIMITER_TIMEOUT", "60"))
 
+REDIS_URL = os.environ.get("REDIS_URL")
+
 RATE_LIMITER_REDIS_OP_TIMEOUT = 5.0
 
 MATCH_ALL = None
@@ -102,7 +104,7 @@ class GlobalRateLimiter:
             dict[tuple[str, str | MatchAllInputs], RateLimitItem] | None
         ) = None,
         use_in_memory: bool = False,
-        redis_url: str | None = None,
+        redis_url: str | None = REDIS_URL,
     ):
         """Initialize a GlobalRateLimiter instance.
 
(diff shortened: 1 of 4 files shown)


packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -30,6 +30,8 @@
 
 GLOBAL_RATE_LIMITER_TIMEOUT = float(os.environ.get("RATE_LIMITER_TIMEOUT", "60"))
```
SamWright822, line 31: I wish we made these global vars instead of magic strings

SamWright822: not this PR, i guess

maykcaldas (approved):

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -113,23 +115,15 @@ def __init__(
...
-                AWS ElastiCache with in-transit encryption), and bare `host:port`
-                (treated as plaintext).
+            redis_url: Redis URL for storage, defaulting to the `REDIS_URL` env var
+                (empty uses in-memory). Must include a scheme: `rediss://` (TLS) or
+                `redis://` (plaintext); a scheme-less URL raises.
         """
         self._rate_config = RATE_CONFIG if rate_config is None else rate_config
-        self._redis_url = redis_url or os.environ.get("REDIS_URL")
+        self._redis_url = redis_url
         self._use_in_memory = use_in_memory or not self._redis_url
```
SamWright822, line 124: Can we add some logging that we are using `_use_in_memory` because no redis_url is set, rather than just using it? 

In the case that use_in_memory is True, but the url is missing, might help with debugging.

packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -30,6 +30,8 @@
 
 GLOBAL_RATE_LIMITER_TIMEOUT = float(os.environ.get("RATE_LIMITER_TIMEOUT", "60"))
 
+REDIS_URL = os.environ.get("REDIS_URL", "")
```
maykcaldas: Do we want this empty string as default? I think we are making changes to always require a url, right?

SamWright822 (commented):
Left a few nit comments, but also I think the readme might be out of date now? 

This section:

```
2. Redis storage (for cross-process rate limiting):

   ```python
   # Set REDIS_URL environment variable
   import os

   os.environ["REDIS_URL"] = "localhost:6379"

   from lmi.rate_limiter import GlobalRateLimiter

   limiter = GlobalRateLimiter()  # Will automatically use Redis if REDIS_URL is set
   ```

can you update that too?

packages/lmi/src/lmi/rate_limiter.py
SamWright822: but this just allows to default to `_use_in_memory` if its blank. I'm not fully sure why we have both - I think we should prob require the redis url always if _use_in_memory is True. 

(written by samw even though i used an em dash)

ypicard: Doesn't change behavior, but defaulting to `None`.

PR #468, opened 2026-07-09
fix(lmi): normalize scheme-less Redis URL instead of rejecting it [CLD-311]

ypicard:
Auto-populate the schema when absent from REDIS_URL.

### DIFF
--- a/packages/lmi/src/lmi/constants.py
+++ b/packages/lmi/src/lmi/constants.py
@@ -6,17 +6,8 @@ _URL_SCHEME_RE = re.compile(r"^[a-z][a-z0-9+.-]*://", re.IGNORECASE)
 
 
 def normalize_redis_url(url: str | None) -> str | None:
-    """Prefix a plaintext `redis://` scheme when the URL has no scheme.
-
-    Surrounding whitespace (e.g. a trailing newline from a secret file) is
-    stripped first; an empty or whitespace-only value normalizes to `None`.
-    """
-    if url is None:
-        return None
-    url = url.strip()
-    if not url:
-        return None
-    if not _URL_SCHEME_RE.match(url):
+    """Prefix a plaintext `redis://` scheme when the URL has no scheme."""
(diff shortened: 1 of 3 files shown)


packages/lmi/src/lmi/rate_limiter.py
```diff
@@ -140,17 +138,6 @@ async def get_outbound_ip(session: aiohttp.ClientSession, url: str) -> str | Non
...
         return None
 
-    @staticmethod
-    def _validate_redis_scheme(redis_url: str | None) -> None:
-        """Require an explicit `redis://` or `rediss://` scheme.
-
-        A scheme-less URL would silently default to plaintext and hang against a
-        TLS-required endpoint (e.g. AWS ElastiCache), so reject it at construction.
-        """
-        if not redis_url or not redis_url.startswith(("redis://", "rediss://")):
```
agolajko, line 150: Is it ok that we are deleting this validation? The new normalistion kinda checks for a slightly different thing

packages/lmi/src/lmi/constants.py
```diff
@@ -1,5 +1,16 @@
+import os
 from sys import version_info
 
+
+def normalize_redis_url(url: str | None) -> str | None:
+    """Ensure a Redis URL carries an explicit scheme."""
+    if url and "://" not in url:
```
agolajko: Are we sure the value passed here won't be something like http://10.58.188.212:6379 wihch doesn't get normalised?

ypicard: If someone passes a value with `http://`, we shouldn't re-write the protocol for them. It's a mistake they should own.

packages/lmi/src/lmi/rate_limiter.py
ypicard: Good point, I can keep it in the normalization layer.

agolajko (approved):

PR #469, opened 2026-07-13
Translate reasoning_effort into OpenRouter's request body at dispatch

elimoss:
Openrouter decided to be all special and use the parameter `reasoning` for reasoning effort, in contrast to what the frontier apis accept, which is `reasoning_effort`. So we have to change it when the model string starts with openrouter.

### DIFF
--- a/packages/lmi/src/lmi/llms.py
+++ b/packages/lmi/src/lmi/llms.py
@@ -899,9 +899,9 @@ def _modify_call_kwargs(call_kwargs: dict[str, Any]) -> dict[str, Any]:
     rejects `reasoning_effort` for any model it has not mapped as
     reasoning-capable, so for OpenRouter models we move a top-level
     `reasoning_effort` into `extra_body.reasoning.effort`, deep-merged into any
-    existing `extra_body` (e.g. a caller-set `provider.only`). Other providers and
-    calls without `reasoning_effort` are returned unchanged. Returns a new dict
-    rather than mutating the input.
+    existing `extra_body` (e.g. a caller-set `provider.only`). Never mutates the
+    input: other providers and calls without `reasoning_effort` get the same dict
+    back unchanged, and the OpenRouter path returns a new dict.
     """
     if (
         not str(call_kwargs.get("model", "")).startswith("openrouter/")
@@ -1484,6 +1484,7 @@ class LiteLLMModel(LLMModel):
         }
         if previous_response_id is not None:
             call_kwargs["previous_response_id"] = previous_response_id
+        call_kwargs = _modify_call_kwargs(call_kwargs)
 
         try:
             response = await track_costs(litellm.aresponses)(**call_kwargs)
@@ -1570,6 +1571,7 @@ class LiteLLMModel(LLMModel):
         }
         if previous_response_id is not None:
             call_kwargs["previous_response_id"] = previous_response_id
+        call_kwargs = _modify_call_kwargs(call_kwargs)
 
         try:
             stream = await track_costs_iter(litellm.aresponses)(**call_kwargs)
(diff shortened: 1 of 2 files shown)


packages/lmi/src/lmi/llms.py
```diff
@@ -890,6 +890,31 @@ def _without_tool_only_kwargs(call_kwargs: dict[str, Any]) -> dict[str, Any]:
...
+    reasoning-capable, so for OpenRouter models we move a top-level
+    `reasoning_effort` into `extra_body.reasoning.effort`, deep-merged into any
+    existing `extra_body` (e.g. a caller-set `provider.only`). Other providers and
+    calls without `reasoning_effort` are returned unchanged. Returns a new dict
+    rather than mutating the input.
+    """
+    if not str(call_kwargs.get("model", "")).startswith("openrouter/"):
+        return call_kwargs
+    if "reasoning_effort" not in call_kwargs:
+        return call_kwargs
```
maykcaldas: I think you can use a `or` to make this a little less verbose

if (
    not startswith openrouter
    or not reasoning_effort in kwargs
):
    return call_kwargs

packages/lmi/src/lmi/llms.py
```diff
@@ -890,6 +890,31 @@ def _without_tool_only_kwargs(call_kwargs: dict[str, Any]) -> dict[str, Any]:
...
+    call_kwargs = dict(call_kwargs)
```
maykcaldas: `call_args` is already a dict. Typed as `dict[str, Any]` Why do we need this cast?

packages/lmi/src/lmi/llms.py
```diff
@@ -890,6 +890,31 @@ def _without_tool_only_kwargs(call_kwargs: dict[str, Any]) -> dict[str, Any]:
...
+    effort = call_kwargs.pop("reasoning_effort")
+    extra_body = dict(call_kwargs.get("extra_body") or {})
+    extra_body["reasoning"] = {**extra_body.get("reasoning", {}), "effort": effort}
```
maykcaldas, line 914: So, just confirming the schema here:
Frontier's APIs use `reasoning_effort` in the kwargs' root.

But openrouter expects this info to be in `kwargs['extra_body']['effort']`?

packages/lmi/src/lmi/llms.py
```diff
@@ -890,6 +890,31 @@ def _without_tool_only_kwargs(call_kwargs: dict[str, Any]) -> dict[str, Any]:
...
+    call_kwargs["extra_body"] = extra_body
+    return call_kwargs
```
maykcaldas, line 916: I believe this is not a stateless function.

Because call_kwargs is passed as a dict and you operate directly over call_kwargs, it'll change call_kwargs in place. 

I'm bringing attention to that because we have the fallback mechanism. If we have a openrouter model and fallback to another provider, this might break

packages/lmi/src/lmi/llms.py
elimoss: `extra_body` is a litellm transport kwarg. the request ends up going out with `kwargs['reasoning']['effort']` set to the reasoning effort. this avoids a break if litellm ever decides to use the same key in its params.

packages/lmi/src/lmi/llms.py
elimoss: that's actually performing a copy that's required to address your comment below about statelessness. since this makes a copy, the original isn't mutated. for clarity i'll change this to a more explicit `.copy()` and add a short comment, as well as a test just to make sure.

packages/lmi/src/lmi/llms.py
elimoss: see previous comment about copying the dict prior to adding the key here

packages/lmi/src/lmi/llms.py
elimoss: heh this was actually what refurb was complaining about in the red ci run

maykcaldas (approved):
it lgtm. 🚀

packages/lmi/src/lmi/llms.py
maykcaldas: Oh. I don't know exactly how dict() copies it. I was expecting to see a deep copy here. Maybe just a copy is enough

PR #471, opened 2026-07-17
Add chat streaming for completions

elimoss:
yield incremental completion tokens as they stream in from the provider followed by a `LLMResult` containing the completed `Message` or `ToolRequestMessage`. The former supports streaming UIs while the latter constitutes the canonical completion. existing calls in non-streaming contexts are unchanged.

also, since we are handling streams now, changed some types around iterators to require aclose so we can cleanly close provider streams when necessary. this is achieved with a `ClosableAsyncIterator` that narrowly requires `aclose`, and `AsyncGenerator` in cases where that is the actual return type


packages/lmi/tests/test_llms.py
```diff
@@ -580,6 +621,407 @@ async def mock_stream() -> AsyncIterator[Any]:  # noqa: RUF029
...
+        with (
+            patch("litellm.acompletion", AsyncMock(return_value=mock_stream())),
+            patch.object(LiteLLMModel, "check_request_limit", request_limit),
+            patch.object(LiteLLMModel, "check_rate_limit", rate_limit),
+        ):
+            stream = await model.call_stream([Message(content="Say hello")])
+            _ = [result async for result in stream]
+
+        assert request_limit.await_count == 1
+        assert rate_limit.await_count == 2
```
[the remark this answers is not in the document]
elimoss: The expected count of 2 includes two distinct checks: one for estimated input tokens before opening the provider stream, and one for the single visible output delta. The terminal result is skipped under `_yield_deltas`, so it is not double-counted. Keeping this assertion at 2 verifies both input and output accounting.

packages/lmi/src/lmi/llms.py
```diff
@@ -371,9 +371,13 @@ async def _commit_stream(gen: AsyncIterable[LLMResult]) -> AsyncIterable[LLMResu
...
     async def replay() -> AsyncIterable[LLMResult]:
-        yield first
-        async for item in iterator:
-            yield item
+        try:
+            yield first
+            async for item in iterator:
+                yield item
+        finally:
+            if aclose := getattr(iterator, "aclose", None):
```
sidnarayanan: Maybe worth a docstring change to note when `aclose` is expected

packages/lmi/src/lmi/llms.py
```diff
@@ -457,6 +461,36 @@ def validate_json_completion(
         ) from err
 
 
+def _tool_parser_accepts_text(parser: Callable[..., Any]) -> bool:
+    """Return whether a custom tool parser accepts completed text."""
```
sidnarayanan: Can you flesh this out? I am not sure what it means for a tool parser to accept completed text vs not. Like maybe provide an example or just write out what's happening - seems like we're checking if the first arg is a string? Also what happens if it's `str | ...`?

sidnarayanan (approved):

packages/lmi/src/lmi/llms.py
elimoss: changed name and docstring to hopefully clarify a bit

packages/lmi/src/lmi/llms.py
elimoss: adding docstring. we can optionally pursue enforcing iterators have `aclose` separately.

packages/lmi/src/lmi/llms.py
```diff
@@ -709,6 +782,80 @@ async def call(  # noqa: C901, PLR0915
...
+        self,
+        messages: list[Message] | str,
+        *,
+        name: str | None = None,
+        tools: list[Tool] | None = None,
+        tool_choice: Tool | str | None = TOOL_CHOICE_REQUIRED,
+        **kwargs: Any,
+    ) -> ClosableAsyncIterator[LLMResult]:
+        """Stream text deltas followed by one canonical completed result."""
+        llm_config = cast("LLMConfig | None", getattr(self, "llm_config", None))
```
sidnarayanan: Hm this feels odd. We have `extra='forbid'` above, so at first glance we shouldn't need a `getattr`. Is this an optional field because of `LiteLLMModel`? And if so, should this implementation be on `LiteLLMModel`?

packages/lmi/src/lmi/llms.py
```diff
@@ -709,6 +782,80 @@ async def call(  # noqa: C901, PLR0915
...
+        chat_kwargs = kwargs.copy()
```
sidnarayanan: Why copy? I don't see `kwargs` consumed elsewhere

packages/lmi/src/lmi/llms.py
```diff
@@ -709,6 +782,80 @@ async def call(  # noqa: C901, PLR0915
...
+                else Message(role=message.role, content=message.content)
+            )
+            for message in messages
+        ]
+        start_clock = asyncio.get_running_loop().time()
+        dispatch_result = await self._run_with_fallbacks(  # type: ignore[attr-defined]
+            self._dispatch,  # type: ignore[attr-defined]
+            messages=messages,
+            streaming=True,
+            _canonical_stream=True,
```
sidnarayanan: What is this? We're routing it through kwargs in `_dispatch`, is that necessary? All these flags (maybe including some not added by these PR) are worthdocstrings.

packages/lmi/src/lmi/llms.py
```diff
@@ -1150,7 +1312,11 @@ async def _dispatch(
 
         # Chat Completions path: `tools` stays in chat_kwargs.
         if streaming:
-            gen = await self.acompletion_iter(messages, spec=spec, **chat_kwargs)
+            gen = await (
+                self._acompletion_stream(messages, spec=spec, **chat_kwargs)
+                if canonical_stream
+                else self.acompletion_iter(messages, spec=spec, **chat_kwargs)
+            )
```
sidnarayanan: Ok I don't quite get this - what are the two different streaming implementations?

packages/lmi/src/lmi/llms.py
```diff
@@ -1458,6 +1633,82 @@ async def acompletion_iter(  # noqa: C901, PLR0915
 
         yield result
 
+    @request_limited
+    @rate_limited
+    async def _acompletion_stream(  # noqa: C901
+        self, messages: list[Message], *, spec: ModelSpec | None = None, **kwargs
+    ) -> AsyncGenerator[LLMResult, None]:
```
sidnarayanan: Let's have docstrings - i.e. what is this vs `acompletion_iter`

sidnarayanan (approved):
Approving to unblock, but can you do a pass on documentation?

elimoss:
got through a prototype and it just wasn't really that much faster in practice.

PR #473, opened 2026-07-20
Require closable async iterators

elimoss:
In a few places we are passing around async iterators that may or may not implement `aclose`. tighten this to specify that they do and add unconditional call to aclose after exit or cancellation. this allows more reliable stream interruption handling and lifecycle management in preparation for https://github.com/Future-House/ldp/pull/471

### DIFF
--- a/packages/lmi/src/lmi/llms.py
+++ b/packages/lmi/src/lmi/llms.py
@@ -377,27 +377,27 @@ class CommonLLMNames(StrEnum):
 
 
 async def _commit_stream(
-    gen: ClosableAsyncIterator[LLMResult],
+    stream: ClosableAsyncIterator[LLMResult],
 ) -> ClosableAsyncIterator[LLMResult]:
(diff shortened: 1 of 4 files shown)


packages/lmi/src/lmi/llms.py
```diff
@@ -356,24 +356,26 @@ class CommonLLMNames(StrEnum):
     )
 
 
-async def _commit_stream(gen: AsyncIterable[LLMResult]) -> AsyncIterable[LLMResult]:
-    """Advance `gen` to its first yield and return an iterator that replays it.
+async def _commit_stream(gen: AsyncGenerator[LLMResult]) -> AsyncGenerator[LLMResult]:
+    """Prime `gen` and return a generator that replays its first result.
```
sidnarayanan: Is "prime" the standard term? The old docstring was more explicit

sidnarayanan (approved):

sidnarayanan (approved):

PR #477, opened 2026-07-22
Test LDP and LMI across supported Python versions

elimoss:
A recent commit caused a break in a dependent codebase in python 3.11 but did not get tested against that version in this repo's ci. add a matrix of python versions so we have more visibility into version-dependent issues before they break dependent code.


sidnarayanan (approved):
eek

